PR #4, opened 2015-01-28
InstallFeature task

oscarlv:
a) The InstallFeature task was added based on the "install-feature" Ant task.
b) Documentation update to add reference to the install feature and improve the format.


jgawor:
Looks good. Thanks!

Issue #5, opened 2015-02-12
The use of the "war" plugin provokes "false positive" in the "deployWar" task.


oscarlv:
I found that in the gradle project is being used the gradle plugin "war" to do the deploy and undeploy of the applications and the use of this plugin is provoking a "false positive" when you call the "deployWar" without specify the params for the deploy.

The params are:
*-destinationDir
-archiveName*

I mean, my build file looks like this:

*File: build.gradle*
```
war{
	destinationDir =file("resourcesToDeploy/")
	archiveName = 'sample.war'
}
```
If you set those params there is not any problem, the result would be successful and the deploy of the application will be done.

![correctsetup](https://cloud.githubusercontent.com/assets/6841943/6176769/09026476-b2c7-11e4-8fec-b65b7bca48ff.PNG)

But, if those params are not set in the build file the result of the "deployWar" task is also successful. Even if has not been specified any file to deploy,  so, I added the following line in the "deployWar" task
just to see which file is being deployed:

*File: Liberty.groovy*
````
project.task('deployWar') {
    doLast {
        println 'File to deploy:' + project.war.archivePath
    }
}
````
(archivePath is the result of concatenate "destinationDir " and "archiveName ") according to the "war" plugin documentation:
http://gradle.org/docs/current/dsl/org.gradle.api.tasks.bundling.War.html

If the parameters are not set:

*File: build.gradle*
```
war{
	//destinationDir =file("resourcesToDeploy/")
	//archiveName = 'sample.war'
}
```

The output (when the parameters weren't set) is this: 

![falsepositive](https://cloud.githubusercontent.com/assets/6841943/6176935/4134e4f8-b2c8-11e4-8a40-9ef21f0f18a8.PNG)

As you can see is looking for a war file with the same name of the project in a folder named "libs" which doesn't exist in my project. 

The structure of my proyect is:
```
-GradlePractices
    -resourcesToDeploy/
    wlp-anttasks-1.1-SNAPSHOT.jar
    liberty-gradle-plugin-1.0-SNAPSHOT.jar
    build.gradle
```

oscarlv:
I've already did a pull request to solve this, please review #9.

oscarlv:
This was solved with #9. Thanks.

PR #6, opened 2015-02-23
Fix issue #5

MParrao:
This code solves the issue #5 also was included the first commit for the integration tests of the plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,15 @@ Of the plugin configuration, only the `wlpDir` property is required. The default
 
 The plugin will have made the following tasks available to your project:
 
-* libertyCreate - Creates a WebSphere Liberty Profile server.
-* libertyStart - Starts the WebSphere Liberty Profile server.
-* libertyRun - Runs a WebSphere Liberty Profile server under the Gradle process.
-* libertyStatus - Checks the WebSphere Liberty Profile server is running.
-* libertyPackage - Generates a WebSphere Liberty Profile server archive.
-* deployWar - Deploys a WAR file to the WebSphere Liberty Profile server.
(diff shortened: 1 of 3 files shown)


jgawor:
A couple of comments:
1) In terms of tests, that's a good start. However, see if there is a way to write an integration tests in such a way that the tests are actually invoking some test.gradle project. See how the integration tests are done in ci.ant or ci.maven

2) Changes to deploy/undeploy are ok but think about supporting multiple applications at the same time. Also naming parameters filePath and fileName is pretty generic and it's hard to see what that will be used for when just looking at a sample .gradle file. Maybe something to consider is "application" or "deployable" type of object for specifying application - see https://github.com/bmuschko/gradle-cargo-plugin as an example.

oscarlv:
@jgawor I've already review what you mentioned but I the "deployable" type it was not defined in the groovy project of the cargo plugin, it was done in the the ant plugin see here - http://cargo.codehaus.org/Ant+support , so to copy this type of object we must to modify the ant tasks to use "deployable" instead of file, fileset, etc. Is that correct?

MParrao:
Maybe this could be done using the file, fileSet and patternSet types directly instead of modifying the deploy/undeploy tasks, also this would make the plugins to be more standarized

MParrao:
I think something like this could help MParrao/ci.gradle@e41b42604729aca5345ec744e36e5429c06c8290  This would allow the user to manipulate better the includes and excludes respective for each task.
For example the user could
```groovy
fileSet = ant.fileset(dir:"src", includes:"**/*.war")
     or
fileSet.setIncludes("**/*.war")
```
Just for giving an example. This will also work with patternset

PR #8, opened 2015-03-23
Add support for installing Liberty from the Liberty repository

MParrao:
a)InstallLiberty task was added to the plugin based on install-liberty
from Ant
b)Parameters were added to the plugin extension
c)Updated documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,8 @@
 .gradle
+.classpath
+.project
+.settings
+bin/
 build
 *~
 /bin/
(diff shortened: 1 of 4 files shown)


jgawor:
Looks pretty good. It would be better to move all these install properties into a separate object. Something like:


    liberty {
        install {
          licenseCode="<license code>"
          runtimeUrl="<url to runtime.jar>"
       }
    }

This would also match closer to what Maven plugin is using.

MParrao:
I made the changes to do this in the way you suggested. Now installFeature task work this way:

`````
liberty {
    installLiberty {
        licenseCode = "<license code>"
        runtimeUrl = " <url to runtime.jar> "
    }
}
````

Please review @0df7048a9b86e2eb9e4c5e6a51bff71ff12734ef

jgawor:
Thanks!

PR #9, opened 2015-03-24
Improve deploy and undeploy tasks.

oscarlv:
This refactor let us to use a closure to deploy/undeploy one or multiple files. 

For example, for multiples files: 
````
liberty {
    deploy {
        dir = "C:/resources"
        include = " *.war, *.ear "
        exclude = " *.txt "    
    }
}
````
Include and Exclude could be set with wildcards.

If you wan to deploy a single file you should use the following code instead:

````
liberty {
    deploy {
        file = "C:/resources/test-war.war"
    }
}
````

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ build
 .classpath
 .project
 .settings/gradle
+
--- a/README.md
+++ b/README.md
@@ -101,8 +101,8 @@ The plugin will have made the following tasks available to your project:
 | libertyDebug | Runs the Liberty Profile server in the console foreground after a debugger connects to the debug port (default: 7777). | 
 | libertyStatus | Checks the WebSphere Liberty Profile server is running. |
 | libertyPackage | Generates a WebSphere Liberty Profile server archive. |
-| deployWar | Deploys a WAR file to the WebSphere Liberty Profile server. |
-| undeployWar | Removes a WAR file from the WebSphere Liberty Profile server. |
+| deploy | Deploys a supported file to the WebSphere Liberty Profile server. |
+| undeploy | Removes an application from the WebSphere Liberty Profile server. |
 | installFeature | Installs a new feature in the WebSphere Liberty Profile server. |
 
 ###Extension properties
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by oscarlv** [comment]
This was updated with the latest changes to the master branch. If anything else is needed here please let me know.

**Comment by jgawor** [comment]
Great. Wondering about two things:
- can you make it work with multiple deploy or undeploy elements? We want to support deploying or undeploying multiple applications.
- is it possible to see if 'war' or 'ear' plugin has been applied to the project and if there is no deploy element specified used the archive location provided by the 'war' / 'ear' project? That way it could seamlessly work with war, ear, or other projects.

**Comment by oscarlv** [comment]
Yes, this could deploy multiple elements if there are in the same dir. I mean, this work as a fileset. For example, to deploy multiple files would be something like this:
```
 liberty {

        wlpDir = 'c:/wlp'
        serverName = 'myServer'

        deploy {
            dir = 'c:/files'
            include = 'app.war, sample.war'
            exclude = 'test-war.war'
        }
    }
```

Is this what you asked?, if not I was wondering if you could reply with an example of how this should work. Thanks.

oscarlv:
This was updated to support deploy using the WAR/EAR plugins. If anything else is needed please let me know.

rmcmx:
@oscarlv Can you please update and resubmit the pull request?

oscarlv:
Done. If anything else is needed please let me know.

rmcmx:
Thanks @oscarlv. I've sent separately some feedback for minor stuff (couple of typos, formatting, etc).

There's two things I think should be handled with higher priority than those though:

- Could you take a look at the changes to README.MD on the commit 2ee6a318823d22836f1c3ca6c1645318e94659e8 ?  Is reporting almost every line as changed.

- In the DeployTask I see the logic implemented to address previous comments. However I think that, if I understood correctly, it should be checked first if the deploy element is specified, and if not, check the war and/or ear elements. As is, I think the code will do backwards: first check war and ear, and then the deploy element. Additionally, I think similar logic should be implemented on the UndeployTask, so if no deploy element is provided, it can work based on ear/war. 

@jgawor Can you please correct me if I'm wrong in this last comment?

oscarlv:
Thanks for the feedback @rmcmx, I've just updated my pull request according to your feedback. Now just is pendig what @jgawor has to say about the logic implemented that you mentioned.

rmcmx:
Thanks @oscarlv for the changes. I added some additional ones and merged.

Issue #13, opened 2015-04-14
The task 'installLiberty' still needs Liberty to already be installed


lazycodeninja:
I was planning to use the installLiberty task to install Liberty as part of a build if it didn't exist, but it seems that the plugin won't load successfully unless the 'bin/tools/ws-server.jar' dependency is defined (which is only available if you've already downloaded Liberty).  Without `classpath files('c:/wlp/bin/tools/ws-server.jar')` the plugin can't be loaded successfully and I get the following stacktrace when running any task (I've cut it down):
```
* Exception is:
org.gradle.api.GradleScriptException: A problem occurred evaluating root project 'liberty-test'.
        at org.gradle.groovy.scripts.internal.DefaultScriptRunnerFactory$ScriptRunnerImpl.run(DefaultScriptRunnerFactory
.java:54)
        ...
Caused by: org.gradle.api.internal.plugins.PluginApplicationException: Failed to apply plugin [id 'liberty']
        at org.gradle.api.internal.plugins.DefaultObjectConfigurationAction.applyPlugin(DefaultObjectConfigurationAction
.java:117)       
        ...
Caused by: org.gradle.api.GradleException: Could not generate a proxy class for class net.wasdev.wlp.gradle.plugins.task
s.RunTask.
        ...
        at net.wasdev.wlp.gradle.plugins.Liberty.apply(Liberty.groovy:47)
        at net.wasdev.wlp.gradle.plugins.Liberty.apply(Liberty.groovy)
        ...
Caused by: java.lang.ClassNotFoundException: com.ibm.wsspi.kernel.embeddable.ServerBuilder
        ... 61 more
```
Looking at the stack trace, the unresolved classes are in the RunTask, which is next to be loaded in Liberty.groovy after the InstallLiberty task.

oscarlv:
Yes, this seems to be a problem with the way the dependencies were defined. Thanks.

lazycodeninja:
Thank you, I've retried with the updated source and it works fine.

PR #14, opened 2015-04-14
Issue13

MParrao:


rmcmx:
Commit merged. Thanks @MParrao for addressing #13

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -44,9 +44,19 @@ class Liberty implements Plugin<Project> {
...
 
-        project.task('libertyRun', type: RunTask) {
-            description = "Runs a WebSphere Liberty Profile server under the Gradle process."
-            logging.level = LogLevel.INFO
+       try {
+            project.task('libertyRun', type: RunTask) {
+                description = "Runs a WebSphere Liberty Profile server under the Gradle process."
+                logging.level = LogLevel.INFO
+            }
+        } catch (Exception e) {
```
StephenTurley: This should never happen. The task should fail if it throws an exception.

PR #27, opened 2015-10-16
Gradleos https://github.com/WASdev/ci.gradle/issues/26

Justoce:
Add "os" parameter to the package server in gradle 
1. Add a new parameter "os"
2. Create a case to set "os" parameter
3. Update documentation


jgawor:
Thanks!

PR #63, opened 2017-06-29
libertyRun functionality and docs

dshimo:
ready for peer review, test not yet ready


src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -57,7 +57,9 @@ class Liberty implements Plugin<Project> {
                 description = "Runs a WebSphere Liberty Profile server under the Gradle process."
                 logging.level = LogLevel.INFO
                 doLast {
-                    println ("This task can't be executed because some dependencies are missing")
+                    logger.error ("libertyRun task could not finish execution due to an exception.")
+                    throw new Exception (e) 
+                    // e.printStackTrace()
```
dnwarnock: Remove debug comment

dshimo: it is removed

docs/libertyRun.md
```diff
@@ -0,0 +1,38 @@
+## libertyRun task  
+Start a Liberty server in foreground. The server instance will be automatically created if it does not exist.
```
cthigh: Hmmm, since we are not using the ant task for run, this second sentence is not true, but we should add it.  If you look at the ant ServerTask.java, you see that the run task starts off by checking:
```
        if (!serverConfigDir.exists()) {
            log(MessageFormat.format(messages.getString("info.server.create"), serverName));
            doCreate();
        }
```
We need to do something similar and call the create task, if the server configuration does not exist.  It should in turn call install, if the runtime does not exist.

docs/libertyRun.md
```diff
@@ -0,0 +1,38 @@
...
+gradle libertyRun --no-daemon
+```
+  
+### 0% Executing?
+While running this task, Gradle will show:  
+```
+<-------------> 0% EXECUTING  
+> :libertyRun
+```  
+This is expected behavior because the task will neither progress nor stop as long as the server/process is running. The "build" will successfully finish with an external `libertyStop` or a Ctrl-C break (if ran with `--no-daemon`).
```
cthigh: Can we remove the caveat "(if ran with --no-daemon)" now?

docs/libertyRun.md
```diff
@@ -0,0 +1,38 @@
...
+
+### What is the Gradle Daemon and why --no-daemon?
+The Gradle Daemon is a long-running background process designed to help speed up the build process. It does so by avoiding constant JVM startup costs while caching information about the project. This behavior is default until specified otherwise.  
+If running with the daemon (default), users who like to Ctrl-C their `libertyRun` task will find that this kills their daemon, ends `libertyRun` prematurely, and most importantly, leaves the server still running. In this state, it's best to restart your server. You will have to execute a separate `libertyStop` command to stop the server before starting it again.
```
cthigh: we need to fix up this sentence, since we are stopping the server now.   How about something along the lines of:
If you are running with the daemon (default), if you stop the `libertyRun` task with a CTRL-C, the daemon stops and you will immediately loose console output. The server will be stopped, but this might take a few seconds longer and you will not receive any feedback on the Gradle console.

docs/libertyRun.md
```diff
@@ -0,0 +1,38 @@
...
+
+### What is the Gradle Daemon and why --no-daemon?
+The Gradle Daemon is a long-running background process designed to help speed up the build process. It does so by avoiding constant JVM startup costs while caching information about the project. This behavior is default until specified otherwise.  
+If running with the daemon (default), users who like to Ctrl-C their `libertyRun` task will find that this kills their daemon, ends `libertyRun` prematurely, and most importantly, leaves the server still running. In this state, it's best to restart your server. You will have to execute a separate `libertyStop` command to stop the server before starting it again.  
+By running `libertyRun` with `--no-daemon`, this task will not benefit from the existing daemon, only suffering a slightly slower startup time. However, this protects the currently running daemon for future tasks and ending the server with a Ctrl-C or an external `libertyStop` command should cleanly end `libertyRun`.
```
cthigh: By running `libertyRun` with `--no-daemon`, you do not get the startup time savings but you do see all messages available from the `libertyRun` task. 

I am not quite sure what you mean with this sentence: However, this protects the currently running daemon for future tasks and ending the server with a Ctrl-C or an external `libertyStop` command should cleanly end `libertyRun`.

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -47,17 +47,24 @@ class Liberty implements Plugin<Project> {
...
+                }
             }
         } catch (Exception e) {
             project.task('libertyRun') {
                 description = "Runs a WebSphere Liberty Profile server under the Gradle process."
                 logging.level = LogLevel.INFO
                 doLast {
-                    println ("This task can't be executed because some dependencies are missing")
+                    logger.error ("libertyRun task could not finish execution due to an exception.")
+                    throw new Exception (e)
```
cthigh: Is it necessary to create a new Exception? Can you just `throw e`?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractTask.groovy
```diff
@@ -70,4 +70,22 @@ abstract class AbstractTask extends DefaultTask {
         return (project.liberty.userDir == null) ? new File(installDir, 'usr') : new File(project.liberty.userDir)
     }
 
+    protected List<String> buildCommand (String operation) {
+        List<String> command = new ArrayList<String>()
+        boolean isWindows = System.properties['os.name'].toLowerCase().indexOf("windows") >= 0
+        String installDir = getInstallDir(project).toString()
+
+        if (isWindows) {
+            // println ("Windows detected")
```
cthigh: take these two println commands out as well.  The if makes it obvious what it is windows.

dshimo: done.

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
dshimo: done

docs/libertyRun.md
dshimo: 1) You're correct: it is not correct to say it has a better startup time but it does have a slightly better command reaction time. The benefit is slight and negligible; maybe it is better to omit the comment.  
  
2) This was just me communicating that using `--no-daemon` may have more long-term use benefits, for example, not setting the build up to constantly kill the current daemon. Should I omit this comment too?

docs/libertyRun.md
dshimo: done.

docs/libertyRun.md
dshimo: done.

docs/libertyRun.md
cthigh: I found it confusing, so maybe so.

docs/libertyRun.md
dshimo: ~done.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
```diff
@@ -21,11 +21,23 @@ class CreateTask extends AbstractTask {
...
+            def params = buildLibertyMap(project);
+            if (project.liberty.template != null && project.liberty.template.length() != 0) {
+                params.put('template', project.liberty.template)
+            }
+            executeServerCommand(project, 'create', params)
+            
+            /*
+            def create_process = new ProcessBuilder(buildCommand("create")).redirectErrorStream(true).start()
+            create_process.inputStream.eachLine {
+                println it
```
dnwarnock: Do we need this commented out code?

dshimo: o still in progress
just checking in changes in case of some freak accident

docs/libertyRun.md
```diff
@@ -0,0 +1,43 @@
+## libertyRun task  
+Start a Liberty server in the foreground. The server instance will be automatically created if it does not exist.  
+**NOTE!** For proper server shutdown and shutdown console messages when terminating `libertyRun` with a Ctrl-C, you **must** run `libertyRun` like this:
```
cthigh: **NOTE:** For proper server shutdown and to view shutdown console messages when terminating `libertyRun` with a Ctrl-C, use the `--no-daemon` option on the `libertyRun` task:

docs/libertyRun.md
```diff
@@ -0,0 +1,43 @@
...
+> :libertyRun
+```  
+or
+```
+<========-----> 66% EXECUTING
+> :libertyRun
+```
+This is expected behavior because the task will neither progress nor stop as long as the server/process is running. The "build" will successfully finish with an external `libertyStop` or a Ctrl-C break.
+
+### What is the Gradle Daemon and why --no-daemon?
```
cthigh: Would it make sense to move this before the section "### 0% or 66% Executing?"

docs/libertyRun.md
```diff
@@ -0,0 +1,43 @@
...
+The Gradle Daemon is a long-running background process designed to help speed up the build process. It does so by caching project information and staying alive, avoiding constant JVM startup costs. This behavior is default until specified otherwise.  
+Gradle's current daemon design makes it difficult to use a `run` task, as a Ctrl-C would kill the daemon while simultaneously leaving the application running in the background. Therefore, we need `--no-daemon` so that the signal can be received and properly handled.  
+In the event that a user does run a `libertyRun` with a daemon (default), an external `libertyStop` must be called in order to properly shut down the server. A Ctrl-C while the Gradle process is running will not stop the server. Use `libertyStatus` to confirm the state of your server.
+
+### Additional parameters
+
+| Parameter | Description | Required |
+| --------- | ------------ | ----------|
+| clean | Setting the `clean` attribute clears all cached information on server start up. It deletes every file in the `${wlp_output_dir}/logs`, `${wlp_output_dir}/workarea`, `${wlp_user_dir}/dropins` or `${wlp_user_dir}/apps`. | No |
```
cthigh: I asked Matt to add a Type column.  This one would be boolean.

cthigh:
Changes look good.  Looks like you have some merge conflicts with Matt's changes.

src/integTest/groovy/net/wasdev/wlp/gradle/plugins/LibertyTest.groovy
```diff
@@ -124,4 +124,22 @@ class LibertyTest extends AbstractIntegrationTest{
            throw new AssertionError ("Fail on task Clean. "+e)
         }
     }
+
+    @Test
+    public void test10_run() {
+        try{
+            def stop_thread = Thread.start {
+                sleep(30000)    // sleep 30 sec
```
cthigh: Faced with a 30 second wait, could this thread check the status every 5 seconds and issue the stop after it detects the server is running?  Then after maybe 30 seconds fail the test if the status never shows running?

docs/libertyRun.md
```diff
@@ -0,0 +1,43 @@
+## libertyRun task
```
cthigh: The table in the main README.md should link to this file as other tasks have done.

PR #66, opened 2017-07-06
Install liberty from maven

josephca:


patricktiu:
We have discussed to use a different configuration property to get the Liberty runtime image from Maven repository (local or Central), just like the Liberty Maven plugin.

For example, (if we are going to use the same property names as the Liberty Maven plugin)
~~~
liberty {
    assemblyArtifact {
          groupId = "com.ibm.websphere.appserver.runtime"
          artifactId = "wlp-webProfile7"
          version = "7.0.0.2"
          type = "zip"
    }
   ...
}
~~~

Your current implementation can only install from remote Maven central.  If the image is already in the local Maven repository cache (file://~/.m2/repsoitory), we should use it instead of downloading it again from Maven Central. 

Other than Maven Central, we should be able to install from other repositories (e.g. sonatype and artifactory).

josephca:
Enabled to use the local repository for maven and updated doc as well. For now, it will look into the default local repository location only.

docs/installLiberty.md
```diff
@@ -1,6 +1,6 @@
...
+The `installLiberty` task is used to download and install WebSphere Liberty server. The task can download the WebSphere Liberty runtime archive from a specified location (via `runtimeUrl`) or automatically resolve it from the [Liberty repository](https://developer.ibm.com/wasdev/downloads/) based on a version and a runtime type, or from the Maven repository. 
 
 In certain cases, the Liberty license code may need to be provided in order to install the runtime. If the license code is required and if you are installing Liberty from the Liberty repository, you can obtain the license code by reading the [current license](https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/17.0.0.2/lafiles/runtime/en.html) and looking for the `D/N: <license code>` line. Otherwise, download the runtime archive and execute `java -jar wlp*runtime.jar --viewLicenseInfo` command and look for the `D/N: <license code>` line.
```
cthigh: Update suggestion since it is tough to keep "current license" up to date.

When installing Liberty from a JAR file, the Liberty license code is needed to install the runtime. When you are installing Liberty from the Liberty repository, you can see the versions of Liberty available to install and find the link to their license using the [index.yml](https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/index.yml) file. After opening the license, look for the `D/N: <license code>` line. Otherwise, download the runtime archive and execute `java -jar wlp*runtime.jar --viewLicenseInfo` command and look for the `D/N: <license code>` line.

docs/installLiberty.md
```diff
@@ -64,3 +64,58 @@ In certain cases, the Liberty license code may need to be provided in order to i
         }
     }
   ```
+
+### Using Maven artifact
+
+Use the `assemblyArtifact` to specify the name of the Maven artifact that contains a custom Liberty server or use one of the provided on the [Maven Central repository](http://search.maven.org/).
```
cthigh: Use the `assemblyArtifact` block

docs/installLiberty.md
```diff
@@ -64,3 +64,58 @@ In certain cases, the Liberty license code may need to be provided in order to i
...
+
+#### assemblyArtifact Properties
+
+These properties are used if `runtimeUrl` is not set.
+
+| Attribute | Description | Required |
+| --------- | ------------ | ----------|
+| groupId | Set Maven groupId to `com.ibm.websphere.appserver.runtime` to  download Liberty runtime archive from the Maven repository. If not set, the [Liberty repository](https://developer.ibm.com/wasdev/downloads/) will be used by default. | No |
+| artifactId | Liberty runtime type to download from the Maven repository. Currently, the following types are supported: `wlp-javaee7`, `wlp-webProfile7`, `wlp-kernel`, `wlp-osgi` and `wlp-microProfile1`. The default value is `wlp-webProfile7`. | No |
+| version | Exact version of the WebSphere Liberty server to install. The default version is '17.0.0.2'. | No |
```
cthigh: Is the default version really 17.0.0.2, or is it the latest?

docs/installLiberty.md
```diff
@@ -64,3 +64,58 @@ In certain cases, the Liberty license code may need to be provided in order to i
...
+
+Note: The group ID for these artifacts is: `com.ibm.websphere.appserver.runtime`.
+
+#### assemblyArtifact Properties
+
+These properties are used if `runtimeUrl` is not set.
+
+| Attribute | Description | Required |
+| --------- | ------------ | ----------|
+| groupId | Set Maven groupId to `com.ibm.websphere.appserver.runtime` to  download Liberty runtime archive from the Maven repository. If not set, the [Liberty repository](https://developer.ibm.com/wasdev/downloads/) will be used by default. | No |
```
cthigh: if they are using maven, shouldn't we default the groupId to `com.ibm.websphere.appserver.runtime`? I am confused why we would default to the liberty repo in this case, isn't the liberty repo used when they configure the `install` block?

src/integTest/resources/liberty-test/install_liberty_javaee7.gradle
```diff
@@ -0,0 +1,37 @@
...
+        mavenCentral()
+        maven {
+            name = 'Sonatype Nexus Snapshots'
+            url = 'https://oss.sonatype.org/content/repositories/snapshots/'
+        }
+    }
+    dependencies {
+        classpath fileTree(dir: '../libs', include: '*.jar', exclude: ['*-groovydoc.jar', '*-sources.jar'])
+        classpath 'net.wasdev.wlp.gradle.plugins:liberty-gradle-plugin:1.1-SNAPSHOT'
+        classpath 'net.wasdev.wlp.ant:wlp-anttasks:1.3-SNAPSHOT'
```
cthigh: Isn't the ant dependency brought in by the liberty-gradle-plugin?

src/integTest/resources/liberty-test/install_liberty_javaee7.gradle
```diff
@@ -0,0 +1,37 @@
...
+    repositories {
+        mavenLocal()
+        mavenCentral()
+        maven {
+            name = 'Sonatype Nexus Snapshots'
+            url = 'https://oss.sonatype.org/content/repositories/snapshots/'
+        }
+    }
+    dependencies {
+        classpath fileTree(dir: '../libs', include: '*.jar', exclude: ['*-groovydoc.jar', '*-sources.jar'])
```
cthigh: what are these dependencies?

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -44,7 +44,7 @@ class Liberty implements Plugin<Project> {
 
         project.task('installLiberty', type: InstallLibertyTask) {
             description 'Installs Liberty from a repository'
-            logging.level = LogLevel.INFO
+            logging.level = LogLevel.DEBUG
```
cthigh: Was this change intentional?  Why?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -30,19 +30,57 @@ class InstallLibertyTask extends AbstractTask {
...
+            type = project.liberty.install.type
         }
 
         if (project.liberty.install.runtimeUrl != null) {
             result.put('runtimeUrl', project.liberty.install.runtimeUrl)
-        }
+        } 
+        else if (project.liberty.assemblyArtifact.groupId != null) {
+            final maven_repo_path = "http://repo1.maven.org/maven2"
+            String version = "17.0.0.2"
```
cthigh: Oh, I see that the version is hard coded.  If we don't specify the version with maven, won't it just get the latest?  I think that would be better than hard coding it to 17.0.0.2.  It is not best practices though.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -30,19 +30,57 @@ class InstallLibertyTask extends AbstractTask {
...
+        if (project.liberty.install.type != null) {
+            result.put('type', project.liberty.install.type)
+            type = project.liberty.install.type
         }
 
         if (project.liberty.install.runtimeUrl != null) {
             result.put('runtimeUrl', project.liberty.install.runtimeUrl)
-        }
+        } 
+        else if (project.liberty.assemblyArtifact.groupId != null) {
```
cthigh: I am still confused with the mix and match of the `install` and the `assemblyArtifact` blocks.  I would expect these to be either/or.

docs/installLiberty.md
```diff
@@ -9,16 +9,16 @@ In certain cases, the Liberty license code may need to be provided in order to i
 | Attribute | Description | Required |
```
cthigh: Are all of these attributes relative to the liberty ` install` block?  We should be clear on that.

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
josephca: Yes, this was intentional. This sets the log level. In order to check debug log which is set by 'logger.debug' in the code, the logging level needs to be lowered to DEBUG

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
josephca: Agreed. I will fix the code to accept either install or assemblyArtifact block to avoid confusion.

src/integTest/resources/liberty-test/install_liberty_javaee7.gradle
josephca: Copied from the sample file - will remove it.

src/integTest/resources/liberty-test/install_liberty_javaee7.gradle
josephca: Correct. Copied from the sample file - will remove it.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
cthigh: I see this is problematic because we are not really downloading using maven coordinates. Instead we are just downloading based on a built URL.

josephca: Will add code to get the latest version from Maven repository.

docs/installLiberty.md
josephca: Changed the code to get the latest version and revised the doc accordingly.

docs/installLiberty.md
josephca: Updated.

docs/installLiberty.md
josephca: Updated.

docs/installLiberty.md
josephca: Updated

docs/installLiberty.md
josephca: Removed the attribute 'groupdId'. I added a note in the doc - specify either 'install' or 'assemblyArtifact' block.

cthigh: The more I think about this, I am not sure it should be removed.  It depends on what you discover related to using gradle dependency management to download the artifact.

josephca: I put it back for the future - in case Maven add additional groupIds.

josephca:
When user specifies 'assemblyArtifact' closure to download Liberty archive from Maven, it will check the file existence in the following orders and use it if found:

- check local Gradle repository
- check local Maven repository
- check remote Maven repository

if the file is not found from above
- download from Liberty repository.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -38,11 +43,79 @@ class InstallLibertyTask extends AbstractTask {
...
+            def ss = project.getConfigurations().getNames()
+            if (!ss.contains('InstallLibertyTaskConfig')) {
+                project.getConfigurations().create('InstallLibertyTaskConfig')
+                project.getDependencies().add('InstallLibertyTaskConfig', groupId + ':' +
+                    artifactId + ':' + version)
+            }
+            
+            String gradleFilePath = project.getConfigurations().getByName('InstallLibertyTaskConfig').getAsPath()
+            logger.debug 'Liberty archive file Path to the local Gradle repository  : ' + 
+                          project.getConfigurations().getByName('InstallLibertyTaskConfig').getAsPath()
```
cthigh: Reuse gradleFilePath variable instead of recalulating the path again.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -38,11 +43,79 @@ class InstallLibertyTask extends AbstractTask {
...
+            }
+            
+            if (project.liberty.assemblyArtifact.type != null) {
+                type = project.liberty.assemblyArtifact.type
+            }
+            
+            def ss = project.getConfigurations().getNames()
+            if (!ss.contains('InstallLibertyTaskConfig')) {
+                project.getConfigurations().create('InstallLibertyTaskConfig')
+                project.getDependencies().add('InstallLibertyTaskConfig', groupId + ':' +
```
cthigh: Does this force the dependency to get automatically downloaded?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
josephca: Good catch! done.

docs/installLiberty.md
```diff
@@ -4,9 +4,9 @@ The `installLiberty` task is used to download and install WebSphere Liberty serv
 
 When installing Liberty from a JAR file, the Liberty license code is needed to install the runtime. When you are installing Liberty from the Liberty repository, you can see the versions of Liberty available to install and find the link to their license using the index.yml file. After opening the license, look for the `D/N: <license code>` line. Otherwise, download the runtime archive and execute `java -jar wlp*runtime.jar --viewLicenseInfo` command and look for the `D/N: <license code>` line.
 
-Note: Either `install` or `assemblyArtifact` closure should be used. They shouldn't be used together. If both are specified the Maven repository with `assemblyArtifact` is used.
+Note: It recommended to use either `install` or `assemblyArtifact` block should be used. If both blocks are specified the Maven repository with `assemblyArtifact` is used.
```
josephca: Updated.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
josephca: Correct. By adding dependency, it will make download the file into the local Gradle repository.

src/integTest/resources/liberty-test/install_liberty_javaee7.gradle
```diff
@@ -0,0 +1,38 @@
...
+
+def test_mode = System.getProperty('runit')
+def wlpLicense = System.getProperty('wlpLicense')
+def wlpVersion = System.getProperty('wlpVersion')
+def WLP_DIR = System.getProperty('wlpInstallDir')
+
+apply plugin: 'liberty'
+
+repositories {
+    mavenCentral()
```
cthigh: Which repositories are being used?  All 3 above or just mavenCentral? Seems like buildscript definitions are passed on to submodules, but I don't quite understand how these resolve.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -15,10 +15,14 @@
...
 package net.wasdev.wlp.gradle.plugins.tasks
 
+import javax.xml.parsers.*
+
 import org.gradle.api.Project
 import org.gradle.api.tasks.TaskAction
 
 class InstallLibertyTask extends AbstractTask {
+    
+    final MAVEN_REPO = 'http://repo1.maven.org/maven2/'
```
cthigh: If adding programmatically adding the dependency causes it to be downloaded, why should we need to reference this URL? What if they are using an internal maven repo?

src/integTest/resources/liberty-test/install_liberty_javaee7.gradle
josephca: According to a gradle doc(https://docs.gradle.org/current/userguide/artifact_dependencies_tutorial.html) 
- A project can have multiple repositories. Gradle will look for a dependency in each repository in the order they are specified, stopping at the first repository that contains the requested module.

josephca: FYI, the 'buildscript -> repositories' is to use external library according to doc - 'If your build script needs to use external libraries, you can add them to the script's classpath in the build script itself.'

The latter 'reopsitories' is for adding dependencies - this is what we're using for this issue.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
josephca: I added the code to download from the maven repositories(internal or remote), in case the file is not downloaded by gradle dependencies programmatically. For instance,  if user doesn't specify maven repository in the gradle script, it won't know about the maven repo. 

- if maven repository is added in the script, it will download Liberty first into the gradle repo then it will use the file from the gradle repo. 
- if maven repository is not added in the script, it will check gradle repository first then maven repositories.

josephca:
More updates are made

- 'version' in assemblyArtifact is required if user want to use Maven repository. If not specified, attributes in assemblyArtifact will be ignored and the Liberty repository will be used instead.
- Liberty runtime file should be available in the specified Maven repositories - mavenLocal() or mavenCentral(). It fails otherwise.

PR #68, opened 2017-07-14
task dependencies in Liberty.groovy

dshimo:


src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -52,8 +52,8 @@ class Liberty implements Plugin<Project> {
             logging.level = LogLevel.INFO
         }
 
-        project.task('libertyStatus', type: StatusTask) {
-            description 'Checks the WebSphere Liberty Profile server is running.'
+        project.task('libertyStatus', type: StatusTask, dependsOn: 'libertyCreate') {
+            description 'Checks if Liberty server is running.'
```
cthigh: if Liberty >>> if the Liberty

PR #69, opened 2017-07-18
Copy config update

mattbsox:
Fixed an issue with the default path values for configuration files

### DIFF
--- a/src/integTest/resources/server-config/src/main/liberty/config/server.xml
+++ b/src/integTest/resources/server-config/src/main/liberty/config/server.xml
@@ -18,7 +18,7 @@
     </featureManager>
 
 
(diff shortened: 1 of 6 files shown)


src/integTest/resources/server-config/src/main/liberty/config/server.xml
```diff
@@ -18,7 +18,7 @@
     </featureManager>
 
 
-    <httpEndpoint httpPort="${default.http.port}" httpsPort="${default.https.port}"
+    <httpEndpoint httpPort="9081" httpsPort="${default.https.port}"
```
dnwarnock: Maybe make this change more obvious so future developer have a better understand as to why the change is here and what the change tests.

dnwarnock:
@cthigh Can we merge? I need it!

PR #70, opened 2017-07-20
Added install-apps task

mattbsox:
Added install-apps task and the ability to specify apps/dropins folder for install-apps.

### DIFF
--- a/src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
+++ b/src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
@@ -33,6 +33,7 @@ import net.wasdev.wlp.gradle.plugins.tasks.InstallFeatureTask
 import net.wasdev.wlp.gradle.plugins.tasks.InstallLibertyTask
 import net.wasdev.wlp.gradle.plugins.tasks.UninstallFeatureTask
 import net.wasdev.wlp.gradle.plugins.tasks.CleanTask
+import net.wasdev.wlp.gradle.plugins.tasks.InstallAppsTask
 
 import org.gradle.api.logging.LogLevel
 
@@ -124,6 +125,11 @@ class Liberty implements Plugin<Project> {
 
         project.task('cleanDirs', type: CleanTask) {
             description 'Deletes files from some directories from the WebSphere Liberty Profile server'
+            logging.level = LogLevel.INFO
+        }
+        
+        project.task('installApps', type: InstallAppsTask) {
+            description "Copy applications specified as either Gradle compile dependencies or the Gradle project package to Liberty server's dropins or apps directory."
             logging.level = LogLevel.INFO
         }
     }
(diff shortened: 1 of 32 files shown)


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -0,0 +1,91 @@
...
+            installProjectArchive()
+        }
+        
+    }
+    
+    
+    private void installProjectArchive() throws Exception {
+        File archive = new File(archivePath())
+        if(!archive.exists()) {
+            throw new GradleException("The project archive was not found and can't be installed.")
```
cthigh: can't >> cannot

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -0,0 +1,91 @@
...
+        }
+        if(project.liberty.installapps.appsDirectory.equals("apps")) {
+            Files.copy(archive.toPath(), new File(getServerDir(project), "/apps/" + archive.getName()).toPath(), StandardCopyOption.REPLACE_EXISTING)
+        }
+        else {
+            Files.copy(archive.toPath(), new File(getServerDir(project), "/dropins/" + archive.getName()).toPath(), StandardCopyOption.REPLACE_EXISTING)
+        }
```
cthigh: you could combine these two and just substitute  project.liberty.installapps.appsDirectory in the Files.copy line.

Files.copy(archive.toPath(), new File(getServerDir(project), "/" + project.liberty.installapps.appsDirectory + "/" + archive.getName()).toPath(), StandardCopyOption.REPLACE_EXISTING)

We might find we need to validate project.liberty.installapps.appsDirectory, but actually it might not be invalid for it to be someplace else, but you can work that out when you do the smart config logic.  It comes down to this: If the app is not configured in the server.xml, it should go in dropins.  If the app is configured in dropins, it is considered relative to the apps folder.  For example, if the location is defined as servlet.war, it would go in apps/servlet.war.  If the location is defined as "foo/servlet.war", it should be put in the apps/foo folder.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -0,0 +1,91 @@
...
+        
+    private String archivePath() {
+        if (project.plugins.hasPlugin("ear")) {
+            return project.ear.archivePath
+        }
+        else if (project.plugins.hasPlugin("war")) {
+            return project.war.archivePath
+        }
+        else {
+            return project.jar.archivePath
```
cthigh: now I see what you mean about using the plugins applied.  Is there a plugin applied to create a jar?  If there, I think put it second and use the war type as the catch all.  Our primary use case is for building war files and they can get the loose application config without the war plugin....

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -116,6 +117,11 @@ class Liberty implements Plugin<Project> {
             description 'Deletes files from some directories from the WebSphere Liberty Profile server'
             logging.level = LogLevel.INFO
         }
+        
+        project.task('installApps', type: InstallAppsTask, dependsOn: 'libertyCreate') {
+            description "Copy applications specified as either Gradle compile dependencies or the Gradle project package to Liberty server's dropins or apps directory."
```
mattbsox: Talked to Cindy. Scratch this comment.

PR #74, opened 2017-07-27
changing licenseArtifact to Gradle style

dshimo:


josephca:
'installLiberty' document and any affected documents should be updated as well.

PR #78, opened 2017-08-02
Added smart configuration functionality to installApps task

mattbsox:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -56,13 +73,37 @@ class InstallAppsTask extends AbstractTask {
...
+        }
+        else if(appsDir.equalsIgnoreCase('dropins') && isAppConfiguredInSourceServerXml(fileName)){
+            throw new GradleException("The application is configured in the server.xml and the plug-in is configured to install the application in the dropins folder. A configured application must be installed to the apps folder.")
         }
-        return project.liberty.installapps.installAppPackages
+    }
+    
+    protected boolean isAppConfiguredInSourceServerXml(String fileName) {
+        boolean configured = false; 
+        if (project.liberty.configFile != null && project.liberty.configFile.exists()) {
```
cthigh: What if we are using configDirectory that includes a server.xml file, does configFile get set?
I couldn't find where it gets set.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -73,6 +114,18 @@ class InstallAppsTask extends AbstractTask {
         return archiveName;
     }
     
+    private String getInstallAppPackages() {
+        if (project.plugins.hasPlugin("ear")) {
+            project.liberty.installapps.installAppPackages = "project"
+        }
```
cthigh: Not sure I understand this method.

src/main/groovy/net/wasdev/wlp/gradle/plugins/utils/ApplicationXmlDocument.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.File;
+import java.io.IOException;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
+
+import org.w3c.dom.Element;
+
+public class ApplicationXmlDocument extends XmlDocument {
+    
+    public static final String APP_XML_FILENAME = "install_apps_configuration_1491924271.xml";
```
cthigh: I never did like this number on this file, but it should probably be different than the Maven one.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -48,6 +52,19 @@ class InstallAppsTask extends AbstractTask {
         if (installProject) {
             installProjectArchive()
         }
+        else if(installDependencies){
+            installDependencies()
```
cthigh: I can't find this method.  What am I missing?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
mattbsox: I'll check. I can add something in to set it.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -56,13 +73,37 @@ class InstallAppsTask extends AbstractTask {
             throw new GradleException("The project archive was not found and cannot be installed.")
         }
         Files.copy(archive.toPath(), new File(getServerDir(project), "/" + project.liberty.installapps.appsDirectory + "/" + getArchiveName(archive.getName())).toPath(), StandardCopyOption.REPLACE_EXISTING)
+    
+        validateAppConfig(getArchiveName(archive.getName()), getBaseName(archive))
```
mattbsox: Maven copies the files before it calls validateAppConfig.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
mattbsox: Good catch. It's not declared or implemented anywhere. Interesting that it lets us do that. I'll take it out for now.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
mattbsox: This is to check if we can ignore the installAppPackages value because we're building an EAR project. Might be a little early to add this in, but it's consistent with the Maven plugin.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
mattbsox: So this depends on whether or not a user has set the `configFile` value along with the `configDirectory`. If it's not set then it will look for the file in the default location. Otherwise, it will get set to the user's value but either way the value won't be used unless the `configDirectory` doesn't have a server.xml file. I think I need to add something for the case where `configFile` isn't set regardless of whether or not the `configDirectory` is being used. It might be a good idea to just get the file from the `serverDir` rather than a different directory because that is the file that will actually be used and we won't need to worry about the values of `configFile` or `configDirectory`. This would have to be called after we copy the files though.

cthigh: I was worried about the file being copied over using configDirectory, but then being referenced by configFile. Maybe we need to talk through this a bit more.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
cthigh: ok - that's fine.  I keep wondering if we really need installAppPackages as a flag.  Something to think more about when we add the install dependencies code.  If the project is an ear or war, we install the result.  If there are dependency applications defined, we install those too.

PR #84, opened 2017-08-17
Server extension update

mattbsox:


src/integTest/resources/sample.servlet/verifyTimeoutSuccessAppsTest.gradle
```diff
@@ -34,7 +34,7 @@ ext {
     testServerHttpsPort = 9443
 
     // This is set in the ibm-web-ext.xml file
-    warContext = 'myLibertyApp'
+    warContext = 'myLibertyApp1'
```
cthigh: I remember that myLibertyApp1 was set in server.xml, but I don't see why this had to change now?  I.e., what is really different?

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -45,7 +47,12 @@ class Liberty implements Plugin<Project> {
         project.configurations.create('libertyLicense')
         project.configurations.create('libertyRuntime')
 
-        def currentServer = project.liberty.server
+        project.afterEvaluate{
+            if (project.liberty.server == null) {
+                project.liberty.server = copyProperties(project.liberty)
+            }
+            setServersForTasks(project)
```
cthigh: Add a comment before this method explaining why you have to set the server object on the tasks.

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -55,92 +62,114 @@ class Liberty implements Plugin<Project> {
...
             description 'Checks if the Liberty server is running.'
             logging.level = LogLevel.INFO
-            server = currentServer
+            server = project.liberty.server
         }
 
         project.task('libertyCreate', type: CreateTask, dependsOn: 'installLiberty') {
             description 'Creates a WebSphere Liberty Profile server.'
-            outputs.file { new File(getUserDir(project), "servers/${project.liberty.serverName}/server.xml") }
+            outputs.file { new File(getUserDir(project), "servers/${project.liberty.server.name}/server.xml") }
```
cthigh: Won't this line be problematic for the old style liberty closure where the object gets set in the afterEvaluate method?  In that case will outputs.file get set to "servers/null/server.xml which will never exist?

cthigh: Can we determine the server name for all combinations and have that in a local variable?

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -55,92 +62,114 @@ class Liberty implements Plugin<Project> {
...
+            }
         }
+        serverMap.remove('class')
+
+        return ServerExtension.newInstance(serverMap)
     }
 
+    void setServersForTasks(Project project){
+        project.tasks.withType(AbstractServerTask).each {task ->
+            task.server = project.liberty.server
```
cthigh: This is going to be problematic when we introduce the named domain objects.  Without some changes here all our named server tasks will end up with the same default server object.  Maybe just add a TODO for now that we need to beware.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractTask.groovy
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corporation 2014, 2017.
+ * (C) Copyright IBM Corporation 2017.
```
cthigh: Actually, this date should stay 2014, 2017.  This is the old class.

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
mattbsox: Ya, this is an issue. We'll need to set this after the project has been evaluated.

PR #88, opened 2017-08-25
Install multiple apps

mattbsox:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -77,14 +93,25 @@ class InstallAppsTask extends AbstractServerTask {
         validateAppConfig(getArchiveName(archive.getName()), getBaseName(archive))
     }
 
-    private void validateAppConfig(String fileName, String artifactId) throws Exception {
-        String appsDir = server.installapps.appsDirectory
+    private void installMultipleApps(List<Task> applications, String appsDir){
+        applications.each{ Task task ->
+            Files.copy(task.archivePath.toPath(), new File(getServerDir(project), "/" + appsDir + "/" + getArchiveName(task.archiveName)).toPath(), StandardCopyOption.REPLACE_EXISTING)
+            validateAppConfig(getArchiveName(task.archiveName), task.baseName, appsDir)
```
jjvilleg: Might need to change task.baseName to getArchiveName(...) since tasks might not be of type war

PR #90, opened 2017-09-01
Reorganize documentation 

cthigh:
Add more task documentation, add server extension to the examples, and work on the properties.


docs/installFeature.md
```diff
@@ -1,43 +1,64 @@
...
+        features {
+            name = ['mongodb-2.0']
+            acceptLicense = true
+        }
+    }
 }
 ```
 
-Also is possible install multiple features in a single closure, for example:
+2. It is possible install multiple features using a single block:
```
dshimo: It is possible install multiple features -> It is possible to install multiple features

docs/libertyRun.md
```diff
@@ -4,7 +4,7 @@ Start a Liberty server in the foreground. The server instance will be automatica
 ```
 gradle libertyRun --no-daemon
 ```
-  
+
 ### What is the Gradle Daemon and why --no-daemon?
```
dshimo: As is, there's several things wrong with `libertyRun`. `--no-daemon` will print verbosely (debug level) and Ctrl-C breaks won't print out all the server-stopping messages. Two temporary fixes include:  
1) Removing `--no-daemon` and documenting that the user must stop the server with an external `libertyStop`
2) Changing `installLiberty`'s debug level back to info in `Liberty.groovy`. This seems to remove about 21-22 lines of output actually (DEBUG output ~831 lines and INFO output about ~809 lines)

Thoughts?

cthigh: What if we said something along the lines of:
Stopping the server with a Ctrl-C break is unreliable from the Gradle command line.  Use the `libertyStop` task to ensure the server is stopped.

cthigh: info seemed like the right log level to me by default, but I was convinced otherwise at one point. I don't remember why.  I am ok going back to info.

docs/installFeature.md
cthigh: Done.

docs/libertyRun.md
dshimo: Joseph made the argument that changing the log level in `Liberty.groovy` set the maximum level of the logger. Therefore it doesn't hurt to give it a higher level as those extra messages won't appear unless specified in the command line.  
  
Changing the logging level does have a direct impact on what appears with a `--no-daemon`, as it makes it more verbose. I agree with your phrasing about the unreliability of a Ctrl-C break. The question now is whether or not we should include `--no-daemon`. 

 `--no-daemon` would require `logger.INFO`, excluding 22 lines of code from an `installLiberty -d`. We could exclude `--no-daemon`, but a Ctrl-C would absolutely force an external `libertyStop`. I personally lean towards the former approach.

docs/libertyCreate.md
```diff
@@ -0,0 +1,43 @@
...
+
+These examples show you different ways to configure servers in your scripts:
+
+1. Configure using `configDirectory`. This is useful when you need to copy multiple files such as `server.xml`, `bootstrap.properties`, and included configuration files.
+
+```groovy
+apply plugin: 'liberty'
+
+liberty {
+    configDirectory = file('src/main/liberty/config')
```
mattbsox: Config file properties can be put in a server extension block.

PR #93, opened 2017-09-07
Task dependencies

jjvilleg:
-Added tests where tasks have different dependencies, for both looseConfig true|false. 
-Changed installApps to get dependencies from the task instead of the project


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -200,8 +200,10 @@ class InstallAppsTask extends AbstractServerTask {
 
     private void addWarEmbeddedLib(Element parent, LooseApplication looseApp, Task task) throws Exception {
       ArrayList<File> deps = new ArrayList<File>();
-      project.configurations.runtime.each { deps.add(it)}
-      project.configurations.providedRuntime.each { deps.remove(it) }
+      task.classpath.each {deps.add(it) println(it)}
+      //Removes WEB-INF/lib/main directory since it is not rquired in the xml
+      if(deps != null && !deps.isEmpty())
+        deps.remove(0)
```
mattbsox: Put this in brackets. There might be a better way to remove that directory from the dependencies.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -200,8 +200,10 @@ class InstallAppsTask extends AbstractServerTask {
 
     private void addWarEmbeddedLib(Element parent, LooseApplication looseApp, Task task) throws Exception {
       ArrayList<File> deps = new ArrayList<File>();
-      project.configurations.runtime.each { deps.add(it)}
-      project.configurations.providedRuntime.each { deps.remove(it) }
+      task.classpath.each {deps.add(it) println(it)}
```
mattbsox: Does this need to be here?

PR #101, opened 2017-09-18
Added configuration in order to publish to the gradle repo. 

jjvilleg:
Pushed to the gradle repo. 

https://plugins.gradle.org/plugin/net.wasdev.wlp.gradle.plugins.Liberty

Please let me know of any wording suggestions you may have. Once everything looks good, I'll finish setting up the jenkins job in order to automatically deploy.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,24 +1,26 @@
 apply plugin: 'groovy'
 apply plugin: 'maven'
 apply plugin: 'signing'
-apply plugin: "com.gradle.plugin-publish"
 
-buildscript{
+
+repositories {
+    mavenCentral()
+    maven {
+        name = 'WASdev Repository'
+        url = 'http://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/maven/repository/'
+    }
+    maven {
+        name = 'Sonatype Nexus Snapshots'
+        url = 'https://oss.sonatype.org/content/repositories/snapshots/'
+    }
+}
+
+buildscript {
   repositories {
-      mavenCentral()
-      maven {
-          name = 'WASdev Repository'
-          url = 'http://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/maven/repository/'
-      }
-      maven {
-          name = 'Sonatype Nexus Snapshots'
-          url = 'https://oss.sonatype.org/content/repositories/snapshots/'
-      }
-      maven {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -74,18 +90,32 @@ task sourcesJar(type: Jar) {
...
-    signing {           
+    signing {
         sign configurations.archives
     }
 }
 
+pluginBundle {
+  website = 'https://github.com/WASdev/ci.gradle'
+  vcsUrl = 'https://github.com/WASdev/ci.gradle'
+  description = 'Gradle plugins for managing Liberty profile servers'
```
cthigh: Remove "profile"

build.gradle
```diff
@@ -74,18 +90,32 @@ task sourcesJar(type: Jar) {
...
+  tags = ['liberty', 'websphere', 'devops']
+
+  plugins {
+    ciGradle {
+      id = 'net.wasdev.wlp.gradle.plugins.Liberty'
+      displayName = 'ci.Gradle'
```
cthigh: Not sure where one would see this name, but wondering if the G should be upper case.  Be consistent.

build.gradle
cthigh: also "plugin" rather than "plugins"

build.gradle
```diff
@@ -23,10 +35,14 @@ dependencies {
     compile localGroovy()
     compile ('net.wasdev.wlp.ant:wlp-anttasks:1.4-SNAPSHOT')
     compile group: 'commons-io', name: 'commons-io', version: '2.5'
+    compile gradleApi()
+    compile localGroovy()
     provided group: 'com.ibm.websphere.appserver.api', name: 'com.ibm.websphere.appserver.spi.kernel.embeddable', version: '1.0.0'
     testCompile 'junit:junit:4.11'
 }
 
+apply plugin: "com.gradle.plugin-publish"
```
cthigh: should this be up with the other plugins?

jjvilleg: Moved to the top with the other plugins.

build.gradle
jjvilleg: Took this from the description on the github page. Went ahead and changed this as well.

build.gradle
jjvilleg: I changed it in both places. For some reason, I thought the G was capitalized in the github page.

PR #109, opened 2017-09-22
simple serverDir exist check for stopping a server

dshimo:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/StopTask.groovy
```diff
@@ -22,7 +22,11 @@ class StopTask extends AbstractServerTask {
     @TaskAction
     void stop() {
         if (new File("${project.buildDir}/wlp/").exists()) {
-            executeServerCommand(project, 'stop', buildLibertyMap(project))
+        	if (getServerDir(project).exists()) {
+            	executeServerCommand(project, 'stop', buildLibertyMap(project))
+        	} else {
+        		logger.error ('There is no server to stop!')
```
cthigh: Making a suggestion for both the new and old message. How about:

There is no server to stop.  The server has not been created.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/StopTask.groovy
```diff
@@ -22,7 +22,11 @@ class StopTask extends AbstractServerTask {
...
+        	}
         } else {
             logger.error ('Server stopped, but the Liberty runtime has not yet been installed!')
         }
```
cthigh: There is no server to stop. The runtime has not been installed.

Issue #107, opened 2017-09-21
add all tasks to a websphere liberty group


scphantm:
In Liberty.groovy, you add the tasks

```
        project.task('installLiberty', type: InstallLibertyTask) {
            description 'Installs Liberty from a repository'
            logging.level = LogLevel.INFO
        }
```

When you don't specify a group, the task is added to the `other` group in the IDE.  In big enterprises like mine where my build system has hundreds of tasks for a particular project and that `other` group gets huge.  if you were to simply change those tasks to something like this

```
        project.task('installLiberty', type: InstallLibertyTask) {
            description 'Installs Liberty from a repository'
            group 'Websphere Liberty'
            logging.level = LogLevel.INFO
        }
```

That would pull those tasks out of `other` and into `Websphere Liberty` in your IDE.  Its a simple change that increases usablity

dshimo:
Done! Thanks

PR #112, opened 2017-09-26
Liberty tasks now put into group 'Liberty'

dshimo:
Instead of falling into 'other' category by default, per request by scphantm


hughesj:
Fixes #107

PR #114, opened 2017-10-02
updated tasks to use WLP_OUTPUT_DIR from server.env

mattbsox:


src/integTest/resources/sample.servlet/testOutputDirs.gradle
```diff
@@ -0,0 +1,116 @@
...
+
+task testWar(type:War){
+    baseName = 'testWar'
+}
+
+task createServerEnv{
+    def serverEnvFile = file("src/main/liberty/config/testOutputDir-server.env")
+    if (!serverEnvFile.exists()) {
+        serverEnvFile.write("WLP_OUTPUT_DIR=${project.buildDir}/testOutputDir")
+    }
```
cthigh: At first, I was wondering if this file could have some residual effect on the next test, but since it isn't names server.env, it should be ok.  In fact, I think this file could just be configured in the source project rather than created dynamically if that is easier.

src/integTest/resources/sample.servlet/testOutputDirs.gradle
```diff
@@ -0,0 +1,116 @@
...
+task printMessageAboutRunningServer {
+    doLast {
+        println "The server is now running at http://localhost:${testServerHttpPort}/${warContext}"
+        println "To stop the server run 'gradle libertyStop'"
+    }
+}
+
+check.dependsOn 'integrationTest'
+libertyCreate.dependsOn 'createServerEnv'
+installApps.dependsOn tasks.withType(War)
```
cthigh: same comments apply from the other build file.

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -205,21 +208,71 @@ class Liberty implements Plugin<Project> {
...
+        Properties envProperties = new Properties()
+        //check etc/server.env and set liberty.outputDir
+        File serverEnvFile = new File(Liberty.getInstallDir(project), 'etc/server.env')
+        if (serverEnvFile.exists()) {
+            envProperties.load(new FileInputStream(serverEnvFile))
+            Liberty.setLibertyOutputDir(project, (String) envProperties.get("WLP_OUTPUT_DIR"))
+        }
+    }
+
+    public static void checkServerEnvProperties(Project project) {
```
cthigh: Go ahead and update this method to take a Server parameter and call it with project.liberty.server. Later, we will be better prepared for multiple servers.

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -205,21 +208,71 @@ class Liberty implements Plugin<Project> {
...
+        }
+    }
+
+    private static void setLibertyOutputDir(Project project, String envOutputDir){
+        if (envOutputDir != null) {
+            project.liberty.outputDir = envOutputDir
+        }
+    }
+
+    private static void setServerOutputDir(Project project, String envOutputDir){
```
cthigh: Add the server parameter.

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -205,21 +208,71 @@ class Liberty implements Plugin<Project> {
...
+        //check etc/server.env and set liberty.outputDir
+        File serverEnvFile = new File(Liberty.getInstallDir(project), 'etc/server.env')
+        if (serverEnvFile.exists()) {
+            envProperties.load(new FileInputStream(serverEnvFile))
+            Liberty.setLibertyOutputDir(project, (String) envProperties.get("WLP_OUTPUT_DIR"))
+        }
+    }
+
+    public static void checkServerEnvProperties(Project project) {
+        Properties envProperties = new Properties()
```
cthigh: I think we only want to do this if thisServer.outputDir is null - not already set by the build file.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractServerTask.groovy
```diff
@@ -73,15 +73,15 @@ abstract class AbstractServerTask extends AbstractTask {
...
 
     protected File getServerDir(Project project){
-        String serverDirectory
-        if(server.outputDir !=null && !server.outputDir.isEmpty()){
-            serverDirectory = server.outputDir
-        }
-        else{
-            serverDirectory = getUserDir(project).toString() + "/servers/" + server.name
-        }
+        return new File(getUserDir(project).toString() + "/servers/" + server.name)
```
cthigh: a lot was removed and I can't remember why.  Can we talk through this one tomorrow?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
```diff
@@ -15,12 +15,15 @@
...
 
+import net.wasdev.wlp.gradle.plugins.Liberty
 import org.gradle.api.tasks.TaskAction
 
 class CreateTask extends AbstractServerTask {
 
     @TaskAction
     void create() {
+        //Checking etc/server.env for outputDir
+        Liberty.checkEtcServerEnvProperties(project)
```
cthigh: was sort of expecting this at the end of liberty Install?

src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -223,18 +223,20 @@ class Liberty implements Plugin<Project> {
...
-    public static void checkServerEnvProperties(Project project) {
+    public static void checkServerEnvProperties(ServerExtension server) {
         Properties envProperties = new Properties()
         //check server.env files and set liberty.server.outputDir
-        if (project.liberty.server.configDirectory != null) {
-            File serverEnvFile = new File(project.liberty.server.configDirectory, 'server.env')
-            if (serverEnvFile.exists()) {
-                envProperties.load(new FileInputStream(serverEnvFile))
-                Liberty.setServerOutputDir(project, (String) envProperties.get("WLP_OUTPUT_DIR"))
+        if (server.outputDir == null) {
```
cthigh: add same check for the etc case... build file takes priority.

PR #118, opened 2017-10-05
added ws-launch.jar condition for if liberty is installed

dshimo:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -24,18 +24,20 @@ class InstallLibertyTask extends AbstractTask {
...
+                                classpath: project.buildscript.configurations.classpath.asPath)
+            project.ant.installLiberty(params)
+
+            String licenseFilePath = project.configurations.getByName('libertyLicense').getAsPath()
+            if (licenseFilePath) {
+                def command = "java -jar " + licenseFilePath + " --acceptLicense " + project.buildDir
+                def process = command.execute()
+                process.waitFor()
+            }
         }
```
cthigh: I think the else of this should say something.  If we called to the ant task, it would say "Liberty profile is already installed."
Maybe we should log.info 
"Liberty is already installed at " + getInstallDir(project)

PR #120, opened 2017-10-05
Install apps fix

mattbsox:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractServerTask.groovy
```diff
@@ -88,15 +88,23 @@ abstract class AbstractServerTask extends AbstractTask {
     protected void setDefaults(Project project){
         if(server.configFile.toString().equals('default')){
             server.configFile = new File(project.projectDir.toString() + '/src/main/liberty/config/server.xml')
+        } else if (server.configFile != null && !server.configFile.exists()) {
```
cthigh: Just realized that since server.configFile is set to "default", it shouldn't ever be null.  You can remove the null check in each of the else clauses.  If it could be null, line 89 would have already blown.

PR #121, opened 2017-10-05
Resources folder now put in the classes directory

jjvilleg:
Loose application will now have src/main/resoures added to WEB-INF/classes/main instead of WEB-INF/lib.

### DIFF
--- a/src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
+++ b/src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
@@ -25,6 +25,7 @@ import groovy.lang.Tuple
 import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.DependencySet
+import org.apache.commons.io.FilenameUtils
 import org.w3c.dom.Element;
 import java.util.regex.Pattern
 import java.util.regex.Matcher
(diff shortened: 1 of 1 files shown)


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -185,8 +186,10 @@ class InstallAppsTask extends AbstractServerTask {
             Element archive = looseApp.addArchive(parent, "/WEB-INF/lib/"+ dep.getName());
             looseApp.addOutputDirectory(archive, project.getRootProject().findProject(projectPath), "/");
             looseApp.addManifestFile(archive, project.getRootProject().findProject(projectPath), "gradle-jar-plugin");
-        } else{
+        } else if(FilenameUtils.getExtension(dep.getAbsolutePath()).equals("jar")){
```
cthigh: maybe we should do an equalsIgnoreCase.  or a toLower() first.

hughesj:
@jjvilleg I'm curious: what problem is this solving? Perhaps this is to bring ci.gradle into parity with ci.maven (not sure if this is what we do in ci.maven). Just looking for some background to this in an issue.

jjvilleg:
@hughesj When we were updating the sample projects, we noticed some projects depended on a persistence.xml that was located in the src/main/resources/META-INF directory.  When Maven installs the app with loose config, the maven-war plugin copies the META-INF directory into /target/classes with the targetInArchive in the loose config being in WEB-INF/classes. Instead of putting it in the build/classes directory , which would be most closely to how Maven does it, the gradle pluging puts it in a build/resources folder. Originally, the loose config would put the META-INF in a WEB-INF/lib/main folder in targetInArchive, but the server was not able to find it. We decided to leave the resources folder as-is, and changed the targetInArchive to be /WEB-INF/classes/main.

PR #135, opened 2017-10-20
Adding plugin-config-xml functionality

mattbsox:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractServerTask.groovy
```diff
@@ -211,6 +215,109 @@ abstract class AbstractServerTask extends AbstractTask {
...
+                    jvmOptionsFile (server.jvmOptionsFile.toString())
+                }
+
+                if (server.serverEnv != null && server.serverEnv.exists()) {
+                    serverEnv (server.serverEnv.toString())
+                }
+
+                appsDirectory (server.appsDirectory)
+                looseApplication (server.looseApplication)
+                stripVersion (server.stripVersion)
```
cthigh: All the server configuration is prime for refactoring to a method that takes a server object and the xml object you need to continue writing to the file. Not sure how that works with this groovy xml writer.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractServerTask.groovy
```diff
@@ -211,6 +215,109 @@ abstract class AbstractServerTask extends AbstractTask {
...
+                installAppPackages ('project')
+
+                if (project.plugins.hasPlugin("war")) {
+                    if (server.looseApplication) {
+                        applicationFileName (project.war.archiveName + '.xml')
+                    } else {
+                        applicationFileName (project.war.archiveName)
```
cthigh: we might need to consider stripVersion here.

PR #136, opened 2017-10-25
method params changed from task to project (LooseApplication)

dshimo:


src/main/groovy/net/wasdev/wlp/gradle/plugins/utils/LooseApplication.groovy
```diff
@@ -36,8 +36,8 @@ public class LooseApplication {
         config.addDir(parent, task.getProject().getAbsolutePath(), target);
     }
 
-    public void addOutputDirectory(Element parent, Task task, String target) {
-        config.addDir(parent, task.getProject().sourceSets.main.getOutput().getClassesDirs().getSingleFile().getAbsolutePath(), target);
+    public void addOutputDirectory(Element parent, Project project, String target) {
```
cthigh: I think it would be better to add the new method and have the old method to call the new.

src/main/groovy/net/wasdev/wlp/gradle/plugins/utils/LooseApplication.groovy
```diff
@@ -47,9 +47,9 @@ public class LooseApplication {
         }
     }
 
-    public void addManifestFile(Element parent, Task task, String pluginId) throws Exception {
+    public void addManifestFile(Element parent, Project project, String pluginId) throws Exception {
```
cthigh: Same here because I don't see any code changes to methods that are calling these existing methods.

src/main/groovy/net/wasdev/wlp/gradle/plugins/utils/LooseApplication.groovy
cthigh: never mind on this comment.

src/main/groovy/net/wasdev/wlp/gradle/plugins/utils/LooseApplication.groovy
cthigh: never mind.

PR #140, opened 2017-10-31
fixed issue with configDirectory

### LINKED ISSUE #139: Config files not copied when configDirectory is specified
When a configDirectory is specified in a build file, the contents of the directory are not being copied to the server directory.

mattbsox:
resolves #139


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractServerTask.groovy
```diff
@@ -129,9 +129,8 @@ abstract class AbstractServerTask extends AbstractTask {
         if (server.configDirectory != null) {
             if(server.configDirectory.exists()){
                 // copy configuration files from configuration directory to server directory if end-user set it
-                def copyAnt = new AntBuilder()
-                copyAnt.copy(todir: serverDirectory) {
-                    fileset(dir: server.configDirectory.getCanonicalPath())
+                server.configDirectory.listFiles().each { file ->
```
cthigh: Do we test nested directories?  Does listFiles go through subfolders or do we need something like walkFileTree?

PR #146, opened 2017-11-03
Resolves XML warning when non-xml document is in configDropins

jjvilleg:


src/main/groovy/net/wasdev/wlp/gradle/plugins/utils/ServerConfigDocument.java
```diff
@@ -236,44 +268,50 @@ private void parseConfigDropinsDir() throws Exception {
...
-            //because we do not want to add locations defined here to the list of locations defined in the other server.xml documents.
-            if (!file.getName().equals(ApplicationXmlDocument.APP_XML_FILENAME)) {
-                parseApplication(doc, "/server/application | /server/webApplication | /server/enterpriseApplication");
+            FileInputStream is = new FileInputStream(file);
+            Document d = parseDocument(is);
+            is.close();
+            return d;
+            } catch (SAXException ex) {
+                //If the file was not valid XML, assume it was some other non XML file in dropins.
+                System.out.println("Dropins file " + file.getAbsolutePath() + " was not parseable as XML");
```
cthigh: Let's make this a log.debug() message.  We generally don't want this message to show up.

Can we change this message to something like:
"Dropins file " + file.getAbsolutePath() + " is not XML and is not evaluated for application configuration."

src/main/groovy/net/wasdev/wlp/gradle/plugins/utils/ServerConfigDocument.java
```diff
@@ -324,18 +362,15 @@ else if (loc.startsWith("ftp:")) {
...
-        }
-        return doc;
+    private Document parseDocument(InputStream in) throws SAXException, IOException {
+         try { //ins will be auto-closed
+            InputStream ins = in;
+            Document d = getDocumentBuilder().parse(ins);
+            ins.close();
+            return d;
+         }
+        catch(Exception e){}
```
cthigh: fix this error - else NPE

PR #151, opened 2017-11-10
Sibling projects now add resources to the loose app

jjvilleg:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -176,10 +177,10 @@ class InstallAppsTask extends AbstractServerTask {
       }
       File parentProjectDir = new File(task.getProject().getRootProject().rootDir.getAbsolutePath())
       for (File dep: deps) {
-        String dependentProjectName = getProjectPath(parentProjectDir, dep)
-        String siblingProject = project.getRootProject().findProject(dependentProjectName)
-        boolean isCurrentProject = ((task.getProject().toString()).contains(dependentProjectName))
-        if ((!isCurrentProject) && !dependentProjectName.isEmpty() && siblingProject != null ){
+        String dependentProjectName = "project ':"+getProjectPath(parentProjectDir, dep)+"'"
```
cthigh: are the single quotes correct before the : and after the project path?

jjvilleg: Yes, this is the format gradle has the task.getProject() property. ex: 
project ':microservice-speaker'

Issue #161, opened 2017-11-23
The libertyCreate task is not copying the modified src server.xml to the server directory


eharris369:
If I modify the src server.xml and then run 'gradle libertyCreate', the changes do not get copied to the server directory in the build folder.

eharris369:
Cindy pointed out the following: You can use the `--rerun-tasks` option to force it and that works. Gradle thinks the task should not run when it should.

hrstoyanov:
Confimed issue with Gradle 4.3.1 on MacOS. Can we get an updated plugin asap, please?

keithchong:
@eharris369, @cthigh , I have a config file that includes another file.   The included file is not being copied over.  Is this a known problem?

mattbsox:
@keithchong If you are just specifying a config file then the included file will not be copied. You can use the configDirectory property to copy the folder that contains both of these files into the server directory.

keithchong:
Thanks @mattbsox .  That's what was missing.  It works, and it also copies the subfolders too.

PR #163, opened 2017-11-27
add input files to be copied over when calling libertyCreate #161

dshimo:
Defines configFile, jvmOptionsFile, bootstrapPropertiesfile, serverEnv as inputs if set or if existing on their default paths. Additionally, configDirectory is added as an input directory if, and only if, set.
#161

### DIFF
--- a/src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
+++ b/src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
@@ -116,12 +116,23 @@ class Liberty implements Plugin<Project> {
                 // Defining files set in build.gradle as inputs
                 if (!project.liberty.server.configFile.toString().equals('default')) {
                     inputs.file { project.liberty.server.configFile }
+                } else if (new File("${projectDir}/src/main/liberty/config/server.xml").exists()) {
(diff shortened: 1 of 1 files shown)


src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -113,6 +113,30 @@ class Liberty implements Plugin<Project> {
...
+                    inputs.file { project.liberty.server.jvmOptionsFile }
+                } else if (new File("${projectDir}/src/main/liberty/config/jvm.options").exists()) {
+                    inputs.file { new File("${projectDir}/src/main/liberty/config/jvm.options") }
+                }
+                if (!project.liberty.server.serverEnv.toString().equals('default')) {
+                    inputs.file { project.liberty.server.serverEnv }
+                } else if (new File("${projectDir}/src/main/liberty/config/server.env").exists()) {
+                    inputs.file { new File("${projectDir}/src/main/liberty/config/server.env") }
+                }
+                if (project.liberty.server.configDirectory.exists()) {
```
cthigh: make sure you try this on a project where configDirectory is not set.  i.e., will this NPE?

Issue #166, opened 2017-11-28
Issue with  looseApplication = false?


hrstoyanov:
I noticed an issue with installing the app as ```looseApplication= false``` (using 2.0.1 and Gradle 4.3.1):
```
server {
         ...
        looseApplication= false
    }
```
Basically ```installApp``` task fails with :
```
java.nio.file.NoSuchFileException: C:\temp\gitlab\recres\recres-webapp\build\libs\recres-webapp-1.0.0-SNAPSHOT.war -> C:\temp\gitlab\recres\recres-webapp\build\wlp\usr\servers\recres_server\apps\recres-webapp.war
```
. The workaround I found was to first deploy with  ```looseApplication= true``` then manually go and delete the generated XML app file in the server, and then switch back to  ```looseApplication= false``` 

Has anyone else seen this issue?

jjvilleg:
This turned out to be an issue with a task output change in gradle 4.3.

hrstoyanov:
Awesome! When can we get a build in maven with this fix in?

cthigh:
The fix is available in the 2.0.2-SNAPSHOT on OSS.  A new point release should be out within a couple weeks.

hrstoyanov:
Thank you! I will use the snapshots for now.

hrstoyanov:
Hm ... I still have the same issue with 2.0.2-SNAPSHOT. Was it put back for some reason?

cthigh:
@jjvilleg, I got an email from @hrstoyanov replying to this thread that he is still seeing this issue on the 2.0.2-SNAPSHOT.  @hrstoyanov, can you make sure you have indeed downloaded the latest snapshot by using the Gradle `--refresh-dependencies` option. We have observed Gradle not downloading the latest snapshot.

hrstoyanov:
@cthigh will re-try it on Monday, where I can run it on a Windows 7 computer. I do not see the issue on Mac, so it maybe Gradle cache issue. Thanks!
Btw, Gradle released 4.4

hrstoyanov:
could not reproduce the issue, please disregard.

hrstoyanov:
@jjvilleg @cthigh @mattbsox 
Well, I spoke too soon... The issue is there even with the latest 2.1-SNAPSHOT and here is some clue:
The problem seems to that **apps** subfolder does not exist in the error message I reported:
```Groovy
[...]

PR #171, opened 2017-11-29
Task output fixes

### LINKED ISSUE #162: The installLiberty task does not install the server properly with Gradle 4.3.1
This might be a problem with Gradle 4.3.1 but I will start here.  Using a simple web app project and doing a 'gradle installLiberty', it creates the wlp directory under the build folder but it is empty.  The same project using an older version of Gradle (4.1) works fine.

mattbsox:
fixes #162 and #166


hrstoyanov:
Is this going to be released on maven central as 2.0.3? I need this fix asap. Thanks.

PR #172, opened 2017-11-30
Facet version minimums

mattbsox:
Setting the minimum version for the jst.web facet to 3.0


src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -52,6 +52,7 @@ class Liberty implements Plugin<Project> {
         //Used to set project facets in Eclipse
         project.pluginManager.apply('eclipse-wtp')
         project.tasks.getByName('eclipseWtpFacet').finalizedBy 'libertyCreate'
+        project.tasks.getByName('eclipseWtpFacet').facet.facet(name:'jst.web', version: '3.0')
```
cthigh: What if there is no jst.web facet already?  Will this create one or cause a NPE? Is this just setting the default to be used later if needed?  What if the project already has the facet defined and it is set to 3.1?  We would not want to downgrade it.

Issue #181, opened 2017-12-07
Warning about depracated  value when plugin is built and tests are ran


jjvilleg:
```
Gradle now uses separate output directories for each JVM language, but this build assumes a single directory for all classes from a source set. This behaviour has been deprecated and is scheduled to be removed in Gradle 5.0
        at org.gradle.api.internal.tasks.DefaultSourceSetOutput.setClassesDir(DefaultSourceSetOutput.java:95)
        at org.gradle.api.internal.tasks.DefaultSourceSetOutput_Decorated.setClassesDir(Unknown Source)
        at org.gradle.api.internal.tasks.DefaultSourceSetOutput.setClassesDir(DefaultSourceSetOutput.java:90)
        at org.gradle.api.internal.tasks.DefaultSourceSetOutput_Decorated.setClassesDir(Unknown Source)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
        at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
        at org.gradle.internal.metaobject.BeanDynamicObject$MetaClassAdapter.invokeMethod(BeanDynamicObject.java:479)
        at org.gradle.internal.metaobject.BeanDynamicObject$MetaClassAdapter.setProperty(BeanDynamicObject.java:358)
        at org.gradle.internal.metaobject.BeanDynamicObject.trySetProperty(BeanDynamicObject.java:176)
        at org.gradle.internal.metaobject.CompositeDynamicObject.trySetProperty(CompositeDynamicObject.java:66)
        at org.gradle.internal.metaobject.AbstractDynamicObject.setProperty(AbstractDynamicObject.java:73)
        at org.gradle.api.internal.tasks.DefaultSourceSetOutput_Decorated.setProperty(Unknown Source)
        at org.codehaus.groovy.runtime.InvokerHelper.setProperty(InvokerHelper.java:197)
        at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.setProperty(ScriptBytecodeAdapter.java:484)
[...]

PR #182, opened 2017-12-08
Resolved Deprecation warning

jjvilleg:
Got rid of the classesDir depricated warning. I register the output directory with the dir() method, and access the directory as a single file since that should be the only output directory in the file collection.


jjvilleg:
Closes #181

Issue #192, opened 2018-01-14
The libertyCreate task always reports UP-TO-DATE after first build


aguibert:
If I modify my server.xml after my gradle project has been built at least once locally, the server.xml in the build directory is not overwritten/updated with the new modifications.

aguibert:
Closing this issue as the fix has been included in 2.1 and confirmed to resolve the issue.

PR #193, opened 2018-01-14
Define inputs and outputs for libertyCreate task

aguibert:
Fixes issue #192

### DIFF
--- a/src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
+++ b/src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
@@ -270,7 +270,7 @@ class Liberty implements Plugin<Project> {
     }
 
     public static void checkEtcServerEnvProperties(Project project) {
-        if (project.liberty.server.outputDir == null) {
+        if (project.liberty.outputDir == null) {
             Properties envProperties = new Properties()
             //check etc/server.env and set liberty.outputDir
             File serverEnvFile = new File(Liberty.getInstallDir(project), 'etc/server.env')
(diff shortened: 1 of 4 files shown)


src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -268,7 +270,7 @@ class Liberty implements Plugin<Project> {
     }
 
     public static void checkEtcServerEnvProperties(Project project) {
-        if (project.liberty.outputDir == null) {
+        if (project.liberty.server.outputDir == null) {
```
mattbsox: We're treating this as a broader liberty level property rather than a server level property, so I think it's best to keep this as `project.liberty.outputDir`. If the user sets an outputDir at the server level it will override whatever is set at the liberty level.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
```diff
@@ -16,10 +16,51 @@
...
+import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.TaskAction
 
+import groovy.lang.Closure;
+
 class CreateTask extends AbstractServerTask {
 
+	@InputDirectory @Optional
+	File getConfigDir() {
+		if(project.liberty.server.configDirectory != null && project.liberty.server.configDirectory)
```
mattbsox: `project.liberty.server` can be changed to `server` for each of the instances in this file.

aguibert:
@mattbsox regarding your comments on the removed `liberty.*` properties, if we do keep these properties in place, they need to be synchronized with the `liberty.server.*` equivalents somehow (which isn't happening currently... I found out the hard way).  
Regarding the backwards compatibility concern, if people are relying on these dead properties still, isn't the plugin completely busted for them?

mattbsox:
Right now we're currently only copying over the properties in the case that there is not any server closure defined. Any build file that uses a liberty closure without a server closure should work. We're not handling the case where properties are separated between the liberty and server closures.

aguibert:
hi @mattbsox I've pushed a new commit which addresses the review comments you've made

PR #207, opened 2018-02-02
Install and create fixes

mattbsox:
Adding reverted fixes back in along with some input/output updates for create and installLiberty

### DIFF
--- a/src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
+++ b/src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
@@ -30,35 +30,68 @@ import groovy.lang.Closure;
 
 class CreateTask extends AbstractServerTask {
 
+    String defaultPath = project.projectDir.toString() + '/src/main/liberty/config/'
+
+    CreateTask() {
+        outputs.upToDateWhen {
+            getServerDir(project).exists() && new File(getServerDir(project), 'server.xml')
+        }
+    }
+
     @InputDirectory @Optional
     File getConfigDir() {
-        if(project.liberty.server.configDirectory != null && project.liberty.server.configDirectory)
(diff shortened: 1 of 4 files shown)


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
```diff
@@ -16,10 +16,79 @@
...
+    }
+
+    @InputFile @Optional
+    File getConfigFile() {
+        if(!server.configFile.toString().equals('default') && server.configFile.exists()) {
+            return server.configFile
+        } else if (server.configDirectory != null && new File(server.configDirectory, 'server.xml').exists()) {
+            return new File(server.configDirectory, 'server.xml')
+        } else if (new File(defaultPath + 'server.xml').exists()) {
+            return new File(project.projectDir.toString() + '/src/main/liberty/config/server.xml')
```
cherylking: Why not use the same `defaultPath + 'server.xml'` like the previous line?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
```diff
@@ -16,10 +16,79 @@
...
+            return new File(server.configDirectory, 'bootstrap.properties')
+        } else if (new File(defaultPath + 'bootstrap.properties').exists()) {
+            return new File(defaultPath + 'bootstrap.properties')
+        }
+    }
+
+    @InputFile @Optional
+    File getJvmOptionsFile() {
+        if (!server.jvmOptionsFile.toString().equals('default') && server.jvmOptionsFile.exists()) {
+            return server.jvmOptionsFile
```
cherylking: After seeing this logic three times now, why not abstract this out into one method and just pass in the file to check and the file name?

PR #208, opened 2018-02-06
Fixing broken unit tests.

scphantm:
I got the loose installer kinda working again for war files at least.  But i have a big problem here.  I can't figure out why this thing won't deploy.

Pull the code, create a `gradle.properties` file and put this in it

`systemProp.gradleTest.include=Liberty-test_Liberty-test-deployGroovyDSLCompatibilitySpec.groovy`

(forgetting this wont hurt anything, you will just run the whole suite instead of that one test)

Then drop to a command and run 

`./gradlew gradleTest`

That will run the integration tests, specifically that one test.  When its done, look at the report.  It states that liberty started normally, states that it attempted to deploy the war file, waited 45 seconds, then died because the message didn't appear in the log file.  I have been going over that task for over an hour now, and i don't see anything wrong.  the output directory is right, the log file has the correct log entries, the war file shows up in the correct folder, server.xml is there, everything is happening except no log entry.  I can't tell if this is an issue with Liberty, the Ant code, or the Gradle code.  But something like 9 of the current 15 failing tests are caused by this issue.  

I can only work on this so sporadically, i can't even remember when it stopped working.  I wanna say when i merged in the IBM master branch before one of my pushes a week or two ago, but i can't be sure.

Can someone here take a look and see if you spot something that i don't see.


scphantm:
its been a week, noone wants to suggest anything?

cherylking:
We had someone looking at it last Friday. We will update as soon as we figure it out.

jjvilleg:
After a bit of investigation, we were able to find the culprit. The `CreateServerDefaultXmlTask` class is generating a default server.xml that isn't installing any profiles. I added ```<feature>jsp-2.3</feature>``` between the featureManager tags, and that got the test passing. Is there a particular reason why you are generating the default server.xml like this instead of using the liberty provided one?

scphantm:
Now that i think for a few minutes, yea, thats exactly whats happening.  The server.xml is an output file in the task, so if gradle sees an output file, but input is blank, it assumes the file was deleted intentionally and removes it from the output.  If you send me a sample of what the default is supposed to look like i can update the task.

jjvilleg:
The default server.xml currently being created by liberty 17.0.0.3 is 
```
<?xml version="1.0" encoding="UTF-8"?>
<server description="new server">

    <!-- Enable features -->
    <featureManager>
        <feature>jsp-2.3</feature>
    </featureManager>

    <!-- To access this server from a remote client add a host attribute to the following element, e.g. host="*" -->
    <httpEndpoint id="defaultHttpEndpoint"
                  httpPort="9080"
                  httpsPort="9443" />
                  
    <!-- Automatically expand WAR files and EAR files -->
    <applicationManager autoExpand="true"/>

</server>
```

Issue #211, opened 2018-02-07
Web facet set to 2.4 instead of 3.0 for simple web project


eharris369:
I have a simple web project and I do a 'gradle build' on it and then import it into eclipse as an existing Gradle project.  If I use the 2.1.1-SNAPSHOT the Dynamic Web Module facet is set to 2.4 instead of 3.0.  If I use the 2.1 release then the Dynamic Web Module facet is set to 3.0 as expected.

cherylking:
Fixed by #212

PR #212, opened 2018-02-07
fixing facet version issue

### LINKED ISSUE #210: Web facet set to 3.0 but EAR facet set to 5.0
If I import a project that has ear and war sub-projects, for the ear project the EAR facet is set to 5.0 (which is JEE5) and for the war project the Dynamic Web Module facet is set to 3.0 (which is JEE6).  I can't find that it is causing any problems but could the EAR facet be set to 6.0 to be consistent with the web facet?

mattbsox:
Fixes #210 and #211

### DIFF
--- a/src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
+++ b/src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
@@ -223,11 +223,9 @@ class Liberty implements Plugin<Project> {
         if(project.plugins.hasPlugin('war')) {
             setFacetVersion(project, 'jst.web', JST_WEB_FACET_VERSION)
         }
-        if(project.plugins.hasPlugin('ear')) {
-            setFacetVersion(project, 'jst.ear', JST_EAR_FACET_VERSION)
-        }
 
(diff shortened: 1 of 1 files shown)


src/main/groovy/net/wasdev/wlp/gradle/plugins/Liberty.groovy
```diff
@@ -220,12 +220,11 @@ class Liberty implements Plugin<Project> {
...
-            if(project.plugins.hasPlugin('war')) {
-                setFacetVersion(project, 'jst.web', JST_WEB_FACET_VERSION)
-            } else if(project.plugins.hasPlugin('ear')) {
-                setFacetVersion(project, 'jst.ear', JST_EAR_FACET_VERSION)
-            }
+        if(project.plugins.hasPlugin('war')) {
+            setFacetVersion(project, 'jst.web', JST_WEB_FACET_VERSION)
+        }
+        if(project.plugins.hasPlugin('ear')) {
+            setFacetVersion(project, 'jst.ear', JST_EAR_FACET_VERSION)
```
cherylking: This is a nit, but why not move line 227 inside the if at line 230?

PR #213, opened 2018-02-09
Added warning message for looseApp support with files. Closes #209

### LINKED ISSUE #209: Enterprise archive in apps directory when looseApplication is true
For the sample.ejb sample, if I update it to use the latest snapshot (2.1.1), it creates an ear in the apps folder of the server instead of a loose application xml file.  In the liberty-plugin-config.xml it has: `<looseApplication>true</looseApplication>` which conflicts with the ear in the apps folder.

I think the problem is with the following in the ejb-ear/build.gradle file:
    apps = [file(ear.archivePath)]
If I change it to this then it creates the xml file as expected:
    apps = [ear]

jjvilleg:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -329,6 +329,9 @@ class InstallAppsTask extends AbstractServerTask {
     protected void installFromFile(File file, String appsDir) {
         Files.copy(file.toPath(), new File(getServerDir(project).toString() + '/' + appsDir + '/' + file.name).toPath(), StandardCopyOption.REPLACE_EXISTING)
         validateAppConfig(file.name, file.name.take(file.name.lastIndexOf('.')), appsDir)
+        if (server.looseApplication) {
+            logger.warn('Application ' + file.getName() + ' was installed as a file as specified. To install as a loose application, specify the plugin or task generting the archive. ')
```
cherylking: fix typo...generating

PR #227, opened 2018-03-05
Install lib output

jjvilleg:
Added another check to account for when different version/zip is specified in build.gradle for the upToDate check.


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -134,4 +135,38 @@ class InstallLibertyTask extends AbstractTask {
             }
         }
     }
+    
+    
+    protected boolean isInstallationUpToDate() {
```
dshimo: consider renaming this method. you can probably pull off at least isInstallUpToDate or installUpToDate. maybe even installUpdateCheck

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -134,4 +135,38 @@ class InstallLibertyTask extends AbstractTask {
...
+        XmlParser pluginXmlParser = new XmlParser()
+        Node libertyPluginConfig = pluginXmlParser.parse(f)
+
+        if(project.configurations.libertyRuntime != null && !project.configurations.libertyRuntime.dependencies.isEmpty()) {
+            if (!libertyPluginConfig.getAt('assemblyArtifact').isEmpty()) {
+                project.configurations.libertyRuntime.dependencies.each { libertyArtifact ->
+                    if(libertyPluginConfig.getAt('assemblyArtifact').getAt('artifactId').text().equals(libertyArtifact.name)
+                        && libertyPluginConfig.getAt('assemblyArtifact').getAt('version').text().equals(libertyArtifact.version)
+                        && libertyPluginConfig.getAt('assemblyArtifact').getAt('groupId').text().equals(libertyArtifact.group)) {
+                            isUpToDate = true;
```
cherylking: So doesn't this mean that if any of the libertyArtifact is the same, it reports that it is up to date regardless of the rest of the libertyArtifacts? It seems like it should check all of them, and only if all of them are the same then it is up to date.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -134,4 +135,38 @@ class InstallLibertyTask extends AbstractTask {
...
+                    }
+                }
+            }
+        }
+        else if (project.liberty.install.runtimeUrl != null) {
+            if (!libertyPluginConfig.getAt('runtimeUrl').isEmpty()) {
+                if( project.liberty.install.runtimeUrl.equals(libertyPluginConfig.getAt('runtimeUrl').text())) {
+                    isUpToDate = true;
```
cherylking: The conditional statements on lines 162, 163 and 164 could all be combined into one.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
jjvilleg: There should only ever be one libertyArtifact specified, but if there is more than one, I believe we are only worried about the first one specified. But agree, logic should change to check only once instead

cherylking:
Can a test be added for this scenario to ensure the update logic is working correctly?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallLibertyTask.groovy
```diff
@@ -73,6 +76,9 @@ class InstallLibertyTask extends AbstractTask {
             result.put('type', project.liberty.install.type)
         }
 
+        if(project.configurations.libertyRuntime.getFiles().size() > 1) {
```
cherylking: Can project.configurations.libertyRuntime be null here?

jjvilleg: This can never be null since we are creating a libertyRuntimeConfiguration in Liberty.groovy.

cherylking: Then why do we check if it is null in several other places in this class?

jjvilleg: I'm guessing they're not necessary, but I'm not opposed to adding the check for consistency

src/integTest/groovy/net/wasdev/wlp/gradle/plugins/AbstractIntegrationTest.groovy
```diff
@@ -78,11 +78,10 @@ abstract class AbstractIntegrationTest {
         GradleConnector gradleConnector = GradleConnector.newConnector()
         gradleConnector.forProjectDirectory(projectDir)
         ProjectConnection connection = gradleConnector.connect()
-
         try {
             BuildLauncher build = connection.newBuild()
             build.setJvmArguments("-DWLP_DIR=$WLP_DIR")
-            build.withArguments("-i");
+            build.withArguments("-Prunit=online");
```
mattbsox: Set this in gradle.properties and we can pass in the property to overwrite it if we want.

src/integTest/resources/liberty-test/build.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
     }
 }
 
-def test_mode = System.getProperty('runit')
+def test_mode = project.findProperty('runit')
```
mattbsox: Change this to `findProperty('runit')` for consistency.

src/integTest/resources/liberty-test/oldStyleTest.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
     }
 }
 
-def test_mode = System.getProperty('runit')
+def test_mode = project.findProperty('runit')
```
mattbsox: Change this to findProperty('runit') for consistency.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
```diff
@@ -59,11 +59,12 @@ class CreateTask extends AbstractServerTask {
     File getServerEnvFile() {
         return getLibertyPropertyFile(server.serverEnv, 'server.env')
     }
-
-    @OutputFile
+    
+    
+    /*@OutputFile
```
mattbsox: Remove this?

cherylking:
This was superseded by #588

PR #230, opened 2018-03-07
moved setEclipseFacets() into the afterEvaluate block

### LINKED ISSUE #228: Facets set to old version for gradle project with 2 web apps
Unzip the attached project archive and then import the multiServer project into Eclipse as an existing Gradle project.  Once it is imported check the Dynamic Web Module facet version on Web1 and Web2 - it is set to 2.4 instead of 3.0.

If I use net.wasdev.wlp.gradle.plugins:liberty-gradle-plugin:2.1 instead of the latest snapshot then the facet versions get set correctly.

[multiServer.zip](https://github.com/WASdev/ci.gradle/files/1786479/multiServer.zip)

mattbsox:
fixes #228


cherylking:
Also, should the new files have copyrights? I see some testcase files do have copyrights, and others do not.

PR #240, opened 2018-03-21
Test revisions

mattbsox:


src/integTest/resources/sample.servlet/installFeatureServerXmlTest.gradle
```diff
@@ -14,7 +14,6 @@ buildscript {
 
 def test_mode = System.getProperty('runit')
```
cherylking: Should this line change like the others to `def test_mode = runit` ?

mattbsox: Ya, missed that. It'll still work the same though. We add the gradle project properties to the system properties so it will still pick it up. I just wanted to use the gradle properties in the build files.

PR #247, opened 2018-04-16
adding check for missing apps and dropins dirs

### LINKED ISSUE #246: installApps doesn't create the apps (or dropins) output folder
If you use the _installApps_ task to install a war file to a server that was created from a server template which doesn't contain empty _apps_ and _dropins_ folders (like the defaultServer template does), then it fails with an exception like the following:

```
Execution failed for task ':MyApp:installApps'.
> java.nio.file.NoSuchFileException: C:\dev\MyApp\build\libs\MyApp.war -> C:\dev\MyApp\build\wlp\usr\servers\MyServer\apps\MyApp.war
```

IMO, the _installApps_ task should create its output folder if necessary instead of relying on it being there in the server (template). I am using version 2.2 of the Liberty Gradle plugin.

mattbsox:
Creating missing apps and dropins folders if server directory was created from a template that does not include these folders.
Fixes #246

### DIFF
--- a/src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
+++ b/src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
@@ -75,7 +75,6 @@ class CreateTask extends AbstractServerTask {
                 params.put('template', server.template)
             }
             executeServerCommand(project, 'create', params)
-            createApplicationFolders()
         }
         copyConfigFiles()
         writeServerPropertiesToXml(project)
@@ -88,20 +87,6 @@ class CreateTask extends AbstractServerTask {
             return new File(server.configDirectory, fileName)
         } else if (new File(DEFAULT_PATH + fileName).exists()) {
             return new File(DEFAULT_PATH + fileName)
-        }
-    }
-
-    void createApplicationFolders() {
-        File serverDir = getServerDir(project)
-        File appsDirectory = new File(serverDir, 'apps')
-        File dropinsDirectory = new File(serverDir, 'dropins')
-
-        if (!appsDirectory.exists()) {
-            appsDirectory.mkdir()
-        }
-
-        if (!dropinsDirectory.exists()) {
-            dropinsDirectory.mkdir()
         }
     }
 }
--- a/src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
+++ b/src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
@@ -75,6 +75,7 @@ class CreateTask extends AbstractServerTask {
                 params.put('template', server.template)
             }
             executeServerCommand(project, 'create', params)
+            createApplicationFolders()
         }
         copyConfigFiles()
         writeServerPropertiesToXml(project)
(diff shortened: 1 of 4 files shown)


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
```diff
@@ -89,4 +90,18 @@ class CreateTask extends AbstractServerTask {
...
         }
     }
+
+    void createApplicationFolders() {
+        File serverDir = getServerDir(project)
+        File appsDirectory = new File(serverDir, 'apps')
+        File dropinsDirectory = new File(serverDir, 'dropins')
+
+        if (!appsDirectory.exists()) {
+            appsDirectory.mkdir()
```
cherylking: If there is a problem creating the directory, does this throw an exception? Or is there a return value that can be checked?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/CreateTask.groovy
```diff
@@ -75,6 +75,7 @@ class CreateTask extends AbstractServerTask {
                 params.put('template', server.template)
             }
             executeServerCommand(project, 'create', params)
+            createApplicationFolders()
```
cherylking: I don't know if it is ok to tie this to server creation. I thought we would going to detect that the apps or dropins folder did not exist during the installApps task instead. Could they run installApps against a server that was not created with our create task?

cherylking: Also, do we need to make similar changes to maven for this?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -378,4 +381,22 @@ class InstallAppsTask extends AbstractServerTask {
...
+
+        try {
+            if (!appsDirectory.exists()) {
+                appsDirectory.mkdir()
+            }
+
+            if (!dropinsDirectory.exists()) {
+                dropinsDirectory.mkdir()
+            }
+        } catch (Exception e) {
```
cherylking: It's bothering me a bit that we attempt to create both the apps folder and dropins folder when they may only need one of them. Do you think the end user would wonder why we created the extra directory?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -322,4 +323,17 @@ class InstallAppsTask extends AbstractServerTask {
...
+    void createApplicationFolder(String appDir) {
+        File serverDir = getServerDir(project)
+        File applicationDirectory = new File(serverDir, appDir)
+
+        try {
+            if (!applicationDirectory.exists()) {
+                applicationDirectory.mkdir()
+            }
+        } catch (Exception e) {
+            throw new GradleException('There was a problem creating the apps or dropins folder in the server directory.', e)
```
cherylking: I think you could use the actual applicationDirectory path in this message instead of saying "apps or dropins".

PR #254, opened 2018-05-21
fixing empty task actions issue

### LINKED ISSUE #253: Possible issue with installLiberty after upgrading to version 2.3?
After upgrading to 2.3  I see the following issue - my custom task no longer install the OL binaries:

```
task customServerInstall {
    dependsOn libertyCreate, deploySharedLib1, deploySharedLib2
}
deploySharedLib1.mustRunAfter libertyCreate
deployOrientSharedLib.mustRunAfter libertyCreate

task deploySharedLib1(type: Copy) {
    description 'Deploy  jars into the OpenLiberty shared libs folder.'
    from configurations.something
    into "$buildDir/wlp/usr/servers/defaultServer/sharedlib1"
}
// ... Same for deploySharedLib2

```
Looking at the reported task execution I see  that the __installLiberty__ task is no longer executed in 2.3! But after adding it to the above list in __customServerInstall__ , I get this:
```
> Task :recres-webapp:installLiberty UP-TO-DATE
Skipping task ':recres-webapp:installLiberty' as it has no actions.
:recres-webapp:installLiberty (Thread[Task worker for ':',5,main]) completed. Took 0.0 secs.
:recres-webapp:customServerInstall (Thread[Task worker for ':',5,main]) started.
```
So, basically now that  __installLiberty__  is enforced, it thinks all is up-to-date, 
but as a result my OL binaries are missing !?

Downgrading to 2.2 fixes the problem, and I don't need to explicitly reference __installLiberty__, since I think __libertyCreate__ forces it. Running the build shows this  (correct) behavior:

```
> Task :recres-webapp:installLiberty
Task ':recres-webapp:installLiberty' is not up-to-date because:
  Task.upToDateWhen is false.
[ant:installLiberty] Offline mode. Using C:\Windows\temp\wlp-cache\openliberty-runtime-18.0.0.1.zip for file:/C:/g/caches/modules-2/files-2.1/io.openliberty/openliberty-runtime/18.0.0.1/a059c422c6ddd53276804b8e6f2ee0b00c97e1a7/openliberty-runtime-18.0.0.1.zip
:recres-webapp:installLiberty (Thread[Task worker for ':' Thread 6,5,main]) completed. Took 35.124 secs.
:recres-webapp:libertyCreate (Thread[Task worker for ':' Thread 6,5,main]) started.
[ant:server] Server defaultServer created.
```

Please advise. My current workaround is to stay with 2.2

mattbsox:
Fixes #253 where tasks were missing actions when referenced as Task objects rather than Strings in a build file.
- Changed how the plugin initializes the tasks
- Changed how multi-server tasks are applied
- Moved some task properties in to the task class files


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/StartTask.groovy
```diff
@@ -26,6 +27,12 @@ class StartTask extends AbstractServerTask {
 
     protected final String START_APP_MESSAGE_REGEXP = "CWWKZ0001I.*"
 
+    StartTask() {
+        description 'Starts the Liberty server.'
```
cherylking: Why are lines 31-33 not inside configure() like so many of the other tasks? Same question for StatusTask.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/PackageTask.groovy
```diff
@@ -16,9 +16,16 @@
 package net.wasdev.wlp.gradle.plugins.tasks
 
 import org.gradle.api.tasks.TaskAction
+import org.gradle.api.logging.LogLevel
 
 class PackageTask extends AbstractServerTask {
 
+    PackageTask() {
+        description 'Generates a Liberty server archive.'
```
cherylking: Same question about configure().

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/JavaDumpTask.groovy
```diff
@@ -16,9 +16,16 @@
 package net.wasdev.wlp.gradle.plugins.tasks
 
 import org.gradle.api.tasks.TaskAction
+import org.gradle.api.logging.LogLevel
 
 class JavaDumpTask extends AbstractServerTask {
 
+    JavaDumpTask() {
+        description 'Dumps diagnostic information from the Liberty server JVM.'
```
cherylking: Same question about configure().

src/main/groovy/net/wasdev/wlp/gradle/plugins/LibertySingleServerTasks.groovy
```diff
@@ -47,140 +28,65 @@ class LibertySingleServerTasks extends LibertyTasks {
...
     void applyTasks() {
-        overwriteTask('compileJSP', CompileJSPTask, {
-            description 'Compile the JSP files in the src/main/webapp directory. '
-            logging.level = LogLevel.INFO
+        project.compileJSP {
             dependsOn 'installLiberty', 'compileJava'
-            group 'Liberty'
-        })
-
-        overwriteTask('installLiberty', InstallLibertyTask, {
```
cherylking: Why did alot of these tasks get removed? installLiberty, libertyStop, libertyDump etc.

mattbsox: They didn't have additional configuration that needed to be added later in the build.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/StartTask.groovy
mattbsox: Ah, looks like I missed a couple. They should be in configure().

src/main/groovy/net/wasdev/wlp/gradle/plugins/LibertySingleServerTasks.groovy
mattbsox: I moved the some properties that were getting set in that file to the individual task files. For some of the tasks those are the only things we were setting so we don't need to configure any additional properties at that point.

PR #267, opened 2018-07-24
Use runtime kernels for install feature tests

ericglau:
For the install feature integration tests, use `openliberty-kernel` and `wlp-kernel` 18.0.0.2 artifacts from Maven Central instead of repackaging the kernel from the runtime zip.  Expand the scope to allow the `wlp` runtime profile to run these tests as well, since Open Liberty features can install to wlp.

This is the gradle equivalent of https://github.com/WASdev/ci.maven/pull/345


ericglau:
Cancelling this PR since the changes are included in #269.

PR #288, opened 2018-12-03
Refactor server configuration to common

chyt:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -244,7 +248,7 @@ class InstallAppsTask extends AbstractServerTask {
         }
         LooseWarApplication looseWar = new LooseWarApplication(task, config)
         looseWar.addSourceDir()
-        looseWar.addOutputDir(looseWar.getDocumentRoot() , task, "/WEB-INF/classes/");
+        looseWar.addOutputDir(looseWar.getDocumentRoot() , task.classpath.getFiles().toArray()[0], "/WEB-INF/classes/");
```
cherylking: Can you explain this change? Maybe add a comment to the code explaining that second parameter.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -282,11 +286,12 @@ class InstallAppsTask extends AbstractServerTask {
             if (resourceTask.getDestinationDir() != null){
                 looseApp.addOutputDir(archive, resourceTask.getDestinationDir(), "/");
             }
-            looseApp.addManifestFile(archive, siblingProject);
+            File manifestFile = project.sourceSets.main.getOutput().getResourcesDir().getParentFile()
```
cherylking: Should there be a ; at the end for consistency?

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
chyt: This was just taken from the refactored `LooseApplication` class. Instead of passing in a task and then resolving the file, I just passed in the file directly. Not sure what exactly that parameter is since I didn't write the original code.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
chyt: Doesn't really matter, Groovy is fine with either.

PR #293, opened 2019-01-16
Missing buildDir fix

mattbsox:
- Creating missing buildDir when installDir is outside buildDir
- Adding new inputs to installLiberty and libertyCreate


mattbsox:
This fixes an issue where the buildDir was not being recreated after a clean when the baseDIr or installDir were outside of the project's buildDir. The cause of this was removing an outputFile for installLiberty which Gradle would use to create the required directories for the outputFile. This resulted in a missing file exception when installLiberty or libertyCreate were being called after a clean.

PR #304, opened 2019-03-04
Add support for installing Spring Boot applications with classifier

anjumfatima90:


src/integTest/groovy/net/wasdev/wlp/gradle/plugins/TestSpringBootApplication.groovy
```diff
@@ -64,7 +64,7 @@ public class TestSpringBootApplication extends AbstractIntegrationTest{
 
 
     @Test
-    public void test_spring_boot_apps() {
+    public void test_spring_boot_apps_20() {
         try {
             addApplToServerXml("src/main/liberty/config/server.xml",
```
mattbsox: Should you should setting the `buildFilename` to the `test_spring_boot_apps_20.gradle` here? I think runTasks() is using `springboot_archive.gradle` up on line 24.

mattbsox: Setting the build.gradle file for each test case might fix the problem you're having later with the other tests when you try to add the app to the server.xml.

Instead of writing to the server.xml file you could also add  different server.xml files for the other builds. This way you wouldn't need to edit them during the tests. Along with the server.xml files, you'd just need to add this to the test build.gradle:
```
liberty {
    server {
        configFile = file('src/main/liberty/config/thisTestServer.xml')
    }    
}
```

src/integTest/resources/sample.springboot/test_spring_boot_apps_with_classifier_15.gradle
```diff
@@ -0,0 +1,44 @@
...
+	mavenCentral()
+}
+dependencies {
+	compile("org.springframework.boot:spring-boot-starter-web:${springBootVersion}")
+	testCompile('org.springframework.boot:spring-boot-starter-test')
+}
+
+liberty {
+    server {
+    install.useOpenLiberty = false
```
mattbsox: I think you can remove this now. Should work on WLP and OL right?

src/integTest/groovy/net/wasdev/wlp/gradle/plugins/TestSpringBootApplication.groovy
```diff
@@ -134,5 +134,44 @@ public class TestSpringBootApplication extends AbstractIntegrationTest{
         }
     }
 
+    //@Test
+    public void test_spring_boot_apps_with_classifier_15() {
+        try {
+            /*
+            addApplToServerXml("src/main/liberty/config/server.xml",
```
mattbsox: Might be able to remove these xml editing methods if you use different server.xml files with different build.gradle files for each of these test builds.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -141,12 +141,34 @@ class InstallAppsTask extends AbstractServerTask {
 
 
     String getArchiveOutputPath() {
-        if (springBootVersion.startsWith('2')) {
-            return project.bootJar.archivePath.getAbsolutePath()
+        String archiveOutputPath;
+
+        if (springBootVersion.startsWith('2.')) {
+            archiveOutputPath = project.bootJar.archivePath.getAbsolutePath()
```
mattbsox: Do we need to check `bootWar' here too? Just wondering since we're checking the war case for 1.5.

mattbsox: Also, since we're checking the archiveOutputPath below, do we even want to check the war case? Not sure how SpringUtil.isSpringUberJar() works with the war case or if it is something we're supporting.

src/integTest/resources/sample.springboot/test_spring_boot_apps_with_classifier_15.gradle
mattbsox: We should add this if we can remove that line.
```
dependencies {
    libertyRuntime group: runtimeGroup, name: runtimeArtifactId, version: runtimeVersion
}
```

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
anjumfatima90: Done. Added bootWar check

src/integTest/groovy/net/wasdev/wlp/gradle/plugins/TestSpringBootApplication.groovy
anjumfatima90: removed

src/integTest/resources/sample.springboot/test_spring_boot_apps_with_classifier_15.gradle
anjumfatima90: done

PR #308, opened 2019-03-15
Add support for Spring Boot JAR and WAR with classifier

anjumfatima90:


src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractServerTask.groovy
```diff
@@ -304,9 +304,6 @@ abstract class AbstractServerTask extends AbstractTask {
     }
 
     protected String getBaseName(Task task) {
-        if (springBootVersion?.startsWith('1')) {
```
cherylking: Not much point in having a method any more when it just returns task.baseName. Is it used many places? If not, I would just remove this method and fix those references.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/InstallAppsTask.groovy
```diff
@@ -103,18 +104,20 @@ class InstallAppsTask extends AbstractServerTask {
...
     private void installMultipleApps(List<Task> applications, String appsDir) {
         applications.each{ Task task ->
-          installProject(task, appsDir)
+            installProject(task, appsDir)
         }
     }
 
     private void installProjectArchive(Task task, String appsDir) {
-        if (task.name == 'bootJar' || task.name == 'bootRepackage') {
+        if("springboot".equals(getPackagingType())) {
```
cherylking: Is it ok that the passed in Task is ignored here? We used to key off of that.

anjumfatima90: The correct task in case of spring boot war case was not getting passed here. So I completely removed the dependency on passed task. Instead, I use task found by findSpringBootTask() method. But the `task` is used in else part.

src/main/groovy/net/wasdev/wlp/gradle/plugins/tasks/AbstractServerTask.groovy
anjumfatima90: Removed

PR #341, opened 2019-10-22
Updating package names and plugin group id

mattbsox:
Updated plugin group id and test build files
Updated package names for plugin code and tests


build.gradle
```diff
@@ -8,9 +8,9 @@ apply plugin: 'maven'
 apply plugin: 'signing'
 apply plugin: "com.gradle.plugin-publish"
 
-archivesBaseName = 'liberty-gradle-plugin'
-group = 'net.wasdev.wlp.gradle.plugins'
-version = '2.7.1-SNAPSHOT'
+archivesBaseName = 'liberty-gradle'
```
cherylking: Shouldn't we keep the same format for the name as maven?

src/main/groovy/io/openliberty/tools/gradle/LibertyTaskFactory.groovy
```diff
@@ -13,29 +13,29 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package net.wasdev.wlp.gradle.plugins
+package io.openliberty.tools.gradle
```
cherylking: update copyright year

PR #342, opened 2019-10-24
Adding default value for serverOutputDir

mattbsox:
Defaulting to server directory for server output directory if not set in `server` or `liberty` extensions.

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
@@ -68,7 +68,7 @@ abstract class AbstractServerTask extends AbstractTask {
         def userDir = getUserDir(project, installDir)
         result.put('userDir', userDir)
 
-        result.put('outputDir', getServerOutputDir(project))
+        result.put('outputDir', getOutputDir(project))
 
         if (server.timeout != null && !server.timeout.isEmpty()) {
             result.put('timeout', server.timeout)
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -103,7 +102,7 @@ abstract class AbstractServerTask extends AbstractTask {
         } else if (project.liberty.outputDir != null) {
             return project.liberty.outputDir
         } else {
-            return getServerDir(project).toString()
+            return getUserDir(project).toString() + "/servers"
```
cherylking: Why would the server output go into the parent "servers" directory?

mattbsox: Yeah, I guess this is kind of confusing. Going to rename this method to getOutputDir(). The ant tasks are deriving the serverOutputDir/serverDirectory from the outputDir we're getting here. I know we mention the default value of outputDir is `${installDir}/usr/servers/${serverName}` in the docs, which it clearly isn't here. It seems the default is coming from the serverDir value ant derives when the task calls init(). We don't explicitly set the value of the outputDir on the Gradle side of things. Since we call getServerOutputDir() when we generate the liberty-plugin-config.xml file for WDT I'll need to update this to use the same value as ant.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -555,9 +551,9 @@ abstract class AbstractServerTask extends AbstractTask {
         serverTask.setInstallDir(installDir)
         serverTask.setUserDir(getUserDir(project, installDir))
 
-        def serverOutputDir = getServerOutputDir(project)
-        if (serverOutputDir != null) {
-            serverTask.setOutputDir(new File(serverOutputDir))
+        def outputDir = getOutputDir(project)
+        if (outputDir != null) {
```
cherylking: How would getOutputDir ever return null now?

mattbsox: It shouldn't. I'll fix that.

PR #352, opened 2019-11-19
Initial dev mode implementation

rzgry:
#349


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,501 @@
...
+                if (dir.equals(sourceDirectory)) {
+                    runGradleTask(gradleBuildLauncher, 'compileJava', 'processResources');
+                }
+
+                if (dir.equals(testSourceDirectory)) {
+                    runGradleTask(gradleBuildLauncher, 'compileTestJava', 'processTestResources');
+                }
+                return true;
+            } catch (BuildException e) {
+                logger.error("Unable to compile");
```
ericglau: Pass the exception too. `logger.error("Unable to compile", e);`

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,501 @@
...
+                return false;
+            } finally {
+                gradleConnection.close();
+            }
+        }
+
+        @Override
+        public void runUnitTests() throws PluginExecutionException, PluginScenarioException {
+            // NOT NEEDED FOR GRADLE ??
```
ericglau: Update the comment since we know it's not needed for gradle

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    private boolean skipTests = false;
+
+    @Option(option = 'skipTests', description = 'Skip tests.')
+    void setSkipTests(boolean skipTests) {
+        this.skipTests = skipTests;
+    }
+
+//    Only calling a single `gradle test` command, no need to have options to
```
ericglau: Clean up or remove the skipUTs and skipITs sections

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,503 @@
...
+            SourceSet mainSourceSet = project.sourceSets.main;
+            SourceSet testSourceSet = project.sourceSets.test;
+
+            sourceDirectory = mainSourceSet.java.srcDirs.iterator().next()
+            testSourceDirectory = testSourceSet.java.srcDirs.iterator().next()
+
+            File outputDirectory = mainSourceSet.java.outputDir;
+            File testOutputDirectory = testSourceSet.java.outputDir;
+            File serverDirectory = getServerDir(project);
+            File configDirectory = new File(project.projectDir, "src/main/liberty/config");
```
ericglau: Can this use `server.configDirectory` instead?

rzgry: It can but I needed to make sure that `server.configDirectory` is set. So I refactored out the logic that sets it in `copyConfigFiles`  in `AbstractServerTask` to an `initializeConfigDirectory` method.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,507 @@
...
+    void setCompileWait(String compileWait) {
+        try {
+            this.compileWait = compileWait.toDouble();
+        } catch (NumberFormatException e) {
+            logger.error(String.format("Unexpected value: %s for dev mode option compileWait. compileWait should be a valid number.", compileWait));
+            throw e;
+        }
+    }
+
+    private int verifyTimeout = 30;
```
mattbsox: Do some of these options map to properties in `ServerExtension`? Do we (or would it make sense to) use those if they get set in build.gradle?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,507 @@
...
+
+            serverTask.setClean(clean);
+
+            return serverTask;
+        }
+
+        @Override
+        public List<String> getArtifacts() {
+            // https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_plugin_and_dependency_management
+            String[] dependencyConfigurationNames = ['compile', 'compileOnly', 'testCompileOnly'];
```
mattbsox: Users could also be using the `api` or `implementation` [configurations](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation). Since you can't access `implementation` directly you should pull the dependency list from `compileClasspath` with `project.configurations.compileClasspath.resolvedConfiguration.resolvedArtifacts.each`. That should cover all the different compile level configurations.

ericglau: @mattbsox According to that document, `implementation` inherits `api` so it may be enough to just add `api` to this list.

mattbsox: Okay, sounds good to me. Try using both of those configurations somewhere in the test cases if possible.

I need to check if we need to update any other parts of the plugin to account for these.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,507 @@
...
+            util.runTestThread(false, executor, -1, false, false);
+            } else {
+                // else watch for key presses immediately
+                util.runHotkeyReaderThread(executor);
+            }
+
+            try {
+                util.watchFiles(buildFile, outputDirectory, testOutputDirectory, executor, artifactPaths, serverXMLFile);
+            } catch (PluginScenarioException e) { // this exception is caught when the server has been stopped by another process
+                logger.info(e.getMessage());
```
cherylking: Shouldn't we provide some context here for this log message?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,507 @@
...
+//            Use the gradle compile task instead of using the DevUtil compile
+//            util.setUseMavenOrGradleCompile(true);
+
+            util.addShutdownHook(executor);
+
+            util.startServer();
+
+            List<String> artifactPaths = util.getArtifacts();
+            File buildFile = project.getBuildFile();
+            File serverXMLFile = getServerXMLFile(server);
```
cherylking: I don't think this getServerXMLFile method is really necessary. The watchFiles method accepts a null for serverXmlFile. If server.serverXmlFile is null, then watchFiles will already monitor the server.xml in the configDirectory. This is what ci.maven does.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
ericglau: @cherylking This message comes from the exception thrown by ci.common's DevUtil.java which already provides the context: `throw new PluginScenarioException("The server has stopped. Exiting dev mode.");`

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
rzgry: Changed to use CLI args if provided, else use the `ServerExtension` properties, else use the defaults values.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -0,0 +1,507 @@
...
+    @Option(option = 'skipTests', description = 'If set to true, do not run any tests in dev mode. The default value is false.')
+    void setSkipTests(boolean skipTests) {
+        this.skipTests = skipTests;
+    }
+
+    boolean libertyDebug = true;
+
+    @Option(option = 'libertyDebug', description = 'Whether to allow attaching a debugger to the running server. The default value is true.')
+    void setLibertyDebug(String libertyDebug) {
+        this.libertyDebug = Boolean.parseBoolean(libertyDebug)
```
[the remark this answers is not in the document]
rzgry: Added some comments as to why it needs to be a String instead of a boolean Option.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
rzgry: Changed to use `compileClasspath` which extends `compile, compileOnly, implementation` and `testCompileClasspath` which extends `testCompile, testCompileOnly, testImplementation`

PR #358, opened 2019-11-26
Dev Mode: Install features after sever.xml change.

rzgry:
#351

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/InstallFeatureTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/InstallFeatureTask.groovy
@@ -40,7 +40,7 @@ class InstallFeatureTask extends AbstractServerTask {
     // temporary serverDir it uses after a change to the server.xml
     private String serverDirectoryParam;
 
-    @Option(option = 'serverDir', description = '(Optional) Server Directory to get the list of features from.')
+    @Option(option = 'serverDir', description = '(Optional) Server directory to get the list of features from.')
     void setServerDirectoryParam(String serverDir) {
         this.serverDirectoryParam = serverDir;
     }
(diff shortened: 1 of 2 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/InstallFeatureTask.groovy
```diff
@@ -110,7 +120,15 @@ class InstallFeatureTask extends AbstractServerTask {
...
+        def serverFeatures = null;
+
+        // if DevMode provides a server directory parameter use that for finding the server features
+        if (serverDirectoryParam != null) {
+            serverFeatures = util.getServerFeatures(new File(serverDirectoryParam));
+        } else if (getServerDir(project).exists()) {
+            serverFeatures = util.getServerFeatures(getServerDir(project));
+        }
 
         Set<String> featuresToInstall = InstallFeatureUtil.combineToSet(pluginListedFeatures, dependencyFeatures, serverFeatures)
```
ericglau: We should have a way to override the first parameter in `util.installFeatures()` to be `true` so that the licenses are accepted.  This is because WebSphere Liberty features need license acceptance.  We could allow another optional command line parameter for `acceptLicense`, and set it to `true` from DevTask.

ericglau: This isn't needed since the build.gradle file should have 
```
liberty {
    server {
        features {
            acceptLicense = true
        }
    }
}
```
 as shown in example 4 of https://github.com/OpenLiberty/ci.gradle/blob/master/docs/installFeature.md#installfeature-task

src/main/groovy/io/openliberty/tools/gradle/tasks/InstallFeatureTask.groovy
```diff
@@ -35,6 +36,15 @@ class InstallFeatureTask extends AbstractServerTask {
         })
     }
 
+    // DevMode uses this option to provide the location of the
+    // temporary serverDir it uses after a change to the server.xml
+    private String serverDirectoryParam;
+
+    @Option(option = 'serverDir', description = '(Optional) Server Directory to get the list of features from.')
```
ericglau: Change description to `(Optional) Server directory containing the configuration files to get the list of features from.`

rzgry: Fixed

Issue #408, opened 2020-01-10
Dev mode integration with MicroShed Testing


aguibert:
Currently Maven dev mode has some integrations with microshed testing which allows the microshed tests to re-use the dev mode server instead of starting a fresh server each time we want to run integration tests in dev mode. We should also have the same integrations for the Gradle dev mode.

The integrations are pretty simple, it's just setting some system properties on the test JVM. Specifically:
- set `microshed_hostname=<liberty server hostname, probably just localhost>`
- set `microshed_http_port=<liberty http port>`
- set `microshed_https_port=<liberty https port>` (only if HTTPS port is configured)

Details on how this is done in the Maven plugin can be found here:
https://github.com/OpenLiberty/ci.maven/blob/0cb2bd1974557e31781120c19a722cdb40fc0b69/liberty-maven-plugin/src/main/java/io/openliberty/tools/maven/server/DevMojo.java#L895-L897

ericglau:
This was addressed in Issue https://github.com/OpenLiberty/ci.gradle/issues/399 with PR https://github.com/OpenLiberty/ci.gradle/pull/404.

By design, [Gradle doesn't propagate system props to the test JVM](https://github.com/gradle/gradle/issues/7773).

In the [libertyDev readme](https://github.com/OpenLiberty/ci.gradle/blob/master/docs/libertyDev.md) we document that the user needs to propagate the properties from the Gradle JVM to the test JVM by using:
```
test {
    systemProperties = System.properties
}
```

They could also propagate specific properties, but so far we have not documented the MicroShed properties (even though they are supported in the implementation in PR https://github.com/OpenLiberty/ci.gradle/pull/404).

@aguibert do you think we should add documentation for enabling specific MicroShed properties?  Or just add a generic comment in the doc directing the user to add the above snippet if they are using MicroShed Testing?

aguibert:
I think propagating all system properties to the test JVM is bad practice because it will override things like `java.class.path` or the `java.user.dir` properties which may not be the same in both instances,

I would prefer that the user doesn't need to configure anything extra to use microshed testing w/ gradle dev mode. Since the gradle plugin controls launching the test JVM, doesn't it have an opportunity to insert whatever system properties it wants? This is how we implemented it on the maven side

ericglau:
Fixed by https://github.com/OpenLiberty/ci.gradle/pull/409

PR #409, opened 2020-01-13
Propogate system properties to the test jvm

rzgry:
Automatically propagate system properties from the gradle jvm to the jvm running tests. #408

### DIFF
--- a/docs/libertyDev.md
+++ b/docs/libertyDev.md
@@ -63,20 +63,3 @@ Tests can read the following system properties to obtain information about the L
 | liberty.http.port | The port used for client HTTP requests. |
 | liberty.https.port | The port used for client HTTP requests secured with SSL (HTTPS). |
 
-In order to properly propagate the system properties from the Gradle JVM running dev mode to the JVM(s) running your tests, you must configure your `build.gradle` to set the system properties for the test JVM(s).
-
-This can be done by setting specific properties for the test JVM.
-```groovy
-test {
-    systemProperty 'liberty.hostname', System.getProperty('liberty.hostname')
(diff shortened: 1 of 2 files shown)


docs/libertyDev.md
```diff
@@ -63,20 +63,3 @@ Tests can read the following system properties to obtain information about the L
 | liberty.http.port | The port used for client HTTP requests. |
 | liberty.https.port | The port used for client HTTP requests secured with SSL (HTTPS). |
 
-In order to properly propagate the system properties from the Gradle JVM running dev mode to the JVM(s) running your tests, you must configure your `build.gradle` to set the system properties for the test JVM(s).
```
cherylking: It might be a good idea to document what properties are being automatically propagated to the test JVM.

ericglau: @cherylking The `microshed` properties are duplicates of the `liberty.*` properties and are used by the MicroShed Testing framework.  They were not intended to be used directly by the user (even though they can be).    The user should use the liberty.* properties instead.
I agree that we can include an example of how to propagate additional properties though.

docs/libertyDev.md
```diff
@@ -63,3 +63,20 @@ Tests can read the following system properties to obtain information about the L
 | liberty.http.port | The port used for client HTTP requests. |
 | liberty.https.port | The port used for client HTTP requests secured with SSL (HTTPS). |
 
+The liberty gradle plugin automatically propagates the system properties in the table above from the Gradle JVM to the JVM(s) running your tests. If you wish to add your own additional system properties you must configure your `build.gradle` file to set the system properties for the test JVM(s).
```
ericglau: `liberty gradle plugin` -> capitalize as `Liberty Gradle plugin`

PR #398, opened 2019-12-13
Relevant tasks should always depend on install features

ericglau:
Fixes https://github.com/OpenLiberty/ci.gradle/issues/395

Some tasks depended on installFeature only if features are configured in build.gradle.  However, features can also be configured in server.xml (or in other xml files that are included in server.xml) and those should also be installed.  

The corresponding tasks should be changed to always depend on installFeature.  installFeature will determine whether features actually need to be installed.

PR #418, opened 2020-01-23
libertyCreate depends on installFeature depends on installLiberty

### LINKED ISSUE #416: Dev mode not installing features after change to server.xml
```
Copied file: /Users/rzegray/Developer/devex/demo-devmode/src/main/liberty/config/server.xml to: /private/var/folders/x6/tc6_9wj567541sczlplbv6jw0000gn/T/tempConfig1945666673317461456/server.xml

> Task :installLiberty UP-TO-DATE
> Task :libertyCreate
[AUDIT   ] CWWKG0016I: Starting server configuration update.
[ERROR   ] CWWKF0001E: A feature definition could not be found for mphealth-2.0
[AUDIT   ] CWWKG0017I: The server configuration was successfully updated in 0.127 seconds.
[AUDIT   ] CWWKF0008I: Feature update completed in 0.112 seconds.

> Task :installFeature
```

ericglau:
Fixes #416

#398 caused installFeature to depend on libertyCreate, and libertyCreate to be finalized by installFeature. So the server is created before the features are installed, which causes the error seen in #416

Instead, the order should be: install Liberty runtime, then install features, then create the server.

So libertyCreate should depend on installFeature, and installFeature should depend on installLiberty.

### DIFF
--- a/docs/installFeature.md
+++ b/docs/installFeature.md
@@ -19,7 +19,7 @@ The `libertyFeature` dependency configuration can install features in Liberty ru
 You need to include `group`, `name`, and `version` values that describes the artifacts to use. An `ext` value for the ESA file type is not required.
 
 ### dependsOn
-`installFeature` depends on `installLiberty`. If no specific features are requested, `installFeature` depends on `libertyCreate` to evaluate the set of features in the server configuration file.
+`installFeature` depends on `installLiberty`.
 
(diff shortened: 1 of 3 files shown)


ericglau:
Cancelling this.  The server needs to be created before you can install features from its server.xml.

PR #433, opened 2020-02-10
Added function to return gradle command example

TrevCraw:


TrevCraw:
Output for gradle plugin is now as follows:

```
> io.openliberty.tools.common.plugins.util.PluginExecutionException: The server has not started within 3 seconds. Consider increasing the server start timeout if this continues to occur. For example, 'gradle libertyDev --serverStartTimeout=90'
```

Issue #434, opened 2020-02-10
Tasks do not show debug messages when running Gradle with -d


ericglau:
Most (or all) tasks in this plugin restrict their own log level to INFO via the line in their source
```
            logging.level = LogLevel.INFO
```

If you run gradle with `-d`, debug messages from the task do not show up.  We should remove the above line from the tasks so that the task respects whatever the user specifies as the log level.

PR #436, opened 2020-02-14
Remove line that set log level from all tasks

TrevCraw:
Fixes https://github.com/OpenLiberty/ci.gradle/issues/434


TrevCraw:
Removing the line that sets the log level resolves the issue of debug lines not showing up. I tested some tasks out to compare regular (non-debug) output with and without the line. It looks like removing the line seems to let a few more output lines through both when no log level parameter is used and when `-i` is used with the gradle command. This seems acceptable to me.

Issue #406, opened 2020-01-09
Java process still using Liberty files after installing features on Windows part 2


turkeylurkey:
The fix for #397 which appeared in OpenLiberty/ci.common#122 must be pulled into this project by updating build.gradle.

cherylking:
The wrong issue was referenced above. It should have been https://github.com/OpenLiberty/ci.common/pull/122

turkeylurkey:
Thank you, I didn't think of typing it that way.

turkeylurkey:
Fixed.

PR #438, opened 2020-02-18
Describe limitation and work around when using Windows, Java 8 and gradle clean

turkeylurkey:
This is a continuation of the work for issue #406.

### DIFF
--- a/docs/clean.md
+++ b/docs/clean.md
@@ -7,6 +7,7 @@ The `cleanDirs` task deletes server specific files including:
   
 Note that the `cleanDirs` task is different from the `clean` task from the Java plugin. `cleanDirs` cleans the server directories whereas `clean` removes the `build` directory. If a server is running during a `clean`, some files may remain and require a second `clean`. Adding a ```clean dependsOn 'libertyStop'``` to your `build.gradle` file can help prevent this. The Liberty `cleanDirs` task automatically depends on the `libertyStop` task.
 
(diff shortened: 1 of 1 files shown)


docs/clean.md
```diff
@@ -7,6 +7,7 @@ The `cleanDirs` task deletes server specific files including:
   
 Note that the `cleanDirs` task is different from the `clean` task from the Java plugin. `cleanDirs` cleans the server directories whereas `clean` removes the `build` directory. If a server is running during a `clean`, some files may remain and require a second `clean`. Adding a ```clean dependsOn 'libertyStop'``` to your `build.gradle` file can help prevent this. The Liberty `cleanDirs` task automatically depends on the `libertyStop` task.
 
+Note also that if you use Java 8 on the Windows operating system beware of the following. Using the `clean` task after using the `installFeatures` or `libertyDev` task may generate an error trying to delete some of the jars in the build directory. If this happens execute `gradle --stop` to stop the daemon and then try again. To avoid the error use the `--no-daemon` option on all your gradle commands.
```
ericglau: The "beware" wording sounds kind of ominous. The description could be rephrased to be shorter and more direct as well. Perhaps something like the following?

```
If you use Java 8 on Windows, note that using the `clean` task after `installFeatures` or `libertyDev` may show an error trying to delete some jars in the build directory. If this happens, run `gradle --stop` to stop the Gradle daemon and then try again. To avoid the error, use the `--no-daemon` option on all of your Gradle commands.
```

turkeylurkey: I agree, this is clearer.

PR #443, opened 2020-03-12
702 properties change

TrevCraw:


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -798,7 +798,8 @@ class DevTask extends AbstractServerTask {
         // which is where the server.xml is located if a specific serverXmlFile
         // configuration parameter is not specified.
         try {
-            util.watchFiles(buildFile, outputDirectory, testOutputDirectory, executor, artifactPaths, serverXMLFile);
+            util.watchFiles(buildFile, outputDirectory, testOutputDirectory, executor, artifactPaths, serverXMLFile,
+                            project.liberty.server.bootstrapPropertiesFile, project.liberty.server.jvmOptionsFile);
```
ericglau: Can this just be simplified to `server.bootstrapPropertiesFile, server.jvmOptionsFile);` ?

cherylking: No it cannot. These are project properties and they are under the liberty.server configuration element in build.gradle. This is the proper name to reference those properties.

PR #445, opened 2020-04-16
Update instructions to use 3.0 release

ericglau:
Update instructions to use 3.0 release instead of 3.0-SNAPSHOT

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'io.openliberty.tools:liberty-gradle-plugin:3.0-SNAPSHOT'
+        classpath 'io.openliberty.tools:liberty-gradle-plugin:3.0'
     }
 }
 ```


README.md
```diff
@@ -45,7 +45,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'io.openliberty.tools:liberty-gradle-plugin:3.0-SNAPSHOT'
+        classpath 'io.openliberty.tools:liberty-gradle-plugin:3.0'
```
cherylking, line 48: This example is specifically showing how to use a snapshot and should not be changed.

cherylking: Maybe you can add a new example of configuring the release version of the plugin, separate from showing how to use a snapshot?

mattbsox: You can also remove the snapshot repo from that config.

ericglau:
Cancelling this for now

PR #454, opened 2020-05-20
Pass the --container option to deploy task and set up ci.common loose app. deploy

turkeylurkey:


src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -274,6 +274,9 @@ class DeployTask extends AbstractServerTask {
         if (outputDir != null && !outputDir.exists() && hasJavaSourceFiles(task.classpath, outputDir)) {
             logger.warn(MessageFormat.format("Installed loose application from project {0}, but the project has not been compiled.", project.name))
         }
+        if (System.getProperty("io.openliberty.tools.gradle.tasks.container")!=null) {
```
ericglau, line 277: We should define this property as a constant somewhere (where it is accessible by both DevTask and DeployTask). Perhaps in AbstractServerTask since both tasks extend from that.

turkeylurkey: Good idea

PR #458, opened 2020-05-29
Add a variable to contain the project directory name

turkeylurkey:
Update the loose app config with the variable name to use and the project directory
Pass the project directory to the common devutil object
Append the variable name and project directory name to bootstrap.properties


src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -60,11 +59,12 @@ class DeployTask extends AbstractServerTask {
         })
     }
 
-    private Boolean container;
+    private Boolean container = null;
 
     @Option(option = 'container', description = 'Run the server in a Docker container instead of locally. The default value is false.')
     void setContainer(boolean container) {
         this.container = container;
+        super.@container = container;
```
ericglau, line 67: If you're setting `super.@container`, why does this file still need its own `private Boolean container` variable?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -159,11 +159,12 @@ class DevTask extends AbstractServerTask {
         this.pollingTest = pollingTest;
     }
 
-    private Boolean container;
+    private Boolean container = null;
 
     @Option(option = 'container', description = 'Run the server in a Docker container instead of locally. The default value is false.')
     void setContainer(boolean container) {
         this.container = container;
+        super.@container = container;
```
ericglau, line 167: If you're setting `super.@container`, why does this file still need its own `private Boolean container` variable?

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
turkeylurkey: Good question: because of DevTask. I'm keeping the implementation the same in DevTask and DeployTask.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
turkeylurkey: In DevTask we have a private class DevTaskUtil:
    private class DevTaskUtil extends DevUtil {
        DevTaskUtil(File serverDirectory, File sourceDirectory, File testSourceDirectory,
        ...
                    boolean libertyDebug, boolean pollingTest, boolean container
        ) throws IOException {
            super(serverDirectory, sourceDirectory, testSourceDirectory, configDirectory, projectDirectory, ....
                    true, true, pollingTest.booleanValue(), **container**.booleanValue());

If we try to use super.@container in this context super refers to DevUtil.class. The Java solution to this is DevTask.super.@container but this results in a very strange runtime failure when you try to run this in the groovy environment. I'm assuming Groovy did not plan for this corner case. If there is a way to do it in Groovy I can implement that. 

Also, I thought the task having a local field was a more natural Groovy idiom and adding the super.@container as an add-on was more in keeping with the overall design.

turkeylurkey: So it is doing it twice for clarity, like saying "This is my dessert, crème brûlée."

PR #465, opened 2020-06-09
Add getContainers() method to access the field in the DevTaskUtil inner class

turkeylurkey:


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -273,7 +277,7 @@ class DevTask extends AbstractServerTask {
         @Override
         public void stopServer() {
             if (container) {
-                // TODO stop the container instead
+                // Shouldn't get here, DevUtil should stop the container instead
```
TrevCraw, line 280: If the code shouldn't get here, should we throw an error instead of just returning?

PR #471, opened 2020-06-18
Update to enable the restart server function

turkeylurkey:
Pass the container option when you redeploy and use the io.openliberty.tools.projectRoot variable name as a Java System property to pass the container option to the liberty create task.
When copying the source files check the Java System property in addition to the container option.
https://github.com/OpenLiberty/ci.maven/issues/800

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
@@ -301,9 +301,10 @@ abstract class AbstractServerTask extends AbstractTask {
             serverEnvPath = server.serverEnvFile.getCanonicalPath()
         }
 
-        if (container != null && container.booleanValue()) {
-            // Set PROJECT_ROOT_NAME so it will be written in config file.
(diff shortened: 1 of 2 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -301,9 +301,10 @@ abstract class AbstractServerTask extends AbstractTask {
             serverEnvPath = server.serverEnvFile.getCanonicalPath()
         }
 
-        if (container != null && container.booleanValue()) {
-            // Set PROJECT_ROOT_NAME so it will be written in config file.
-            server.var."io.openliberty.tools.projectRoot" = project.getProjectDir().getAbsolutePath()
+        if (container != null && container.booleanValue() ||
+            System.getProperty(PROJECT_ROOT_NAME) != null) { // property set by CreateTask
```
ericglau, line 305: Is the comment `// property set by CreateTask` inaccurate? Since the property is not actually set by CreateTask, but it's set by DevTask when calling CreateTask.

PR #477, opened 2020-06-24
Merge devModeContainers branch to master

ericglau:
Part of https://github.com/OpenLiberty/ci.maven/issues/659

Required for https://github.com/OpenLiberty/ci.common/pull/155 

This is the Gradle equivalent of https://github.com/OpenLiberty/ci.maven/pull/858


.travis.yml
```diff
@@ -10,7 +10,7 @@ env:
     - RUNTIME=ol RUNTIME_VERSION=19.0.0.9
 before_install:
     - echo 'Installing ci.common lib ....'
-    - git clone https://github.com/OpenLiberty/ci.common.git ./ci.common
+    - git clone -b devModeContainers https://github.com/OpenLiberty/ci.common.git ./ci.common
```
kevinwlau, line 13: If this development branch is not going to stay around, reminder to update this to point back to the default branch after these dependencies are finalized and merged. 

Same for appveyor.yml, and same files in ci.maven repo - 4 files in total

src/main/groovy/io/openliberty/tools/gradle/tasks/DevcTask.groovy
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.openliberty.tools.gradle.tasks
+
+import org.gradle.api.tasks.TaskAction
+
+class DevcTask extends DevTask {
+
```
mattbsox, line 21: Can we add a description for the task?

Something along these lines should be good.
```
    DevcTask() {
        configure({
            description 'Runs a Liberty server in dev mode inside of a Docker container'
            group 'Liberty'
        })
    }
```

mattbsox: Looks like the current description is coming from DevTask.

ericglau: Fixed

.travis.yml
ericglau: Reverted

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -158,6 +159,39 @@ class DevTask extends AbstractServerTask {
         this.pollingTest = pollingTest;
     }
 
+    private Boolean container = null;
+
+    @Option(option = 'container', description = 'Run the server in a Docker container instead of locally. The default value is false for the libertyDev task, and true for the libertyDevc task.')
```
cherylking, line 164: I'm very confused why this `private Boolean container` is defined on 3 different tasks so far. And then this is the second time I have seen `@Option(option = 'container'` with a `setContainer` method.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -48,6 +48,7 @@ class DeployTask extends AbstractServerTask {
 
     protected ApplicationXmlDocument applicationXml = new ApplicationXmlDocument();
     private static final boolean DEFAULT_CONTAINER = false;
+    private static final String PROJECT_ROOT_BUILD_LIBS = "build/libs";
```
cherylking, line 51: This just occurred to me (and it may apply to maven as well), but shouldn't we use the project build dir instead of hardcoding "build/"? For gradle that is `project.buildDir`.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
turkeylurkey: `@Option(option = 'container'` can decorate the private variable but it can also decorate the `setContainer` method. We use the set method to pass the option to the super class for use in a method there. 
It is defined multiple times because the dev task calls the deploy task and needs to pass the `container` option. The subtask e.g. deploy runs in a separate JVM so it would seem an option is the clearest way to pass the value.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -60,6 +62,14 @@ class DeployTask extends AbstractServerTask {
         })
     }
 
+    private Boolean container = null;
+
+    @Option(option = 'container', description = 'Run the server in a Docker container instead of locally. The default value is false.')
+    void setContainer(boolean container) {
```
cherylking, line 68: Is this container option available for someone not using the `devc` goal to set and deploy to a container? Would that work?

src/main/groovy/io/openliberty/tools/gradle/extensions/LibertyExtension.groovy
```diff
@@ -38,6 +38,8 @@ class LibertyExtension {
 
     ServerExtension server = server = new ServerExtension()
 
+    DevExtension dev = new DevExtension();
```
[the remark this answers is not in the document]
ericglau: The expected pattern when using containers is one Liberty runtime with one server with one app.  We can make it as `liberty -> server -> dev` but that seems unnecessary.  In the [design document](https://ibm.box.com/s/e5pva5w2shjjzz14w2kz3m7nzjbol87r) slide 25, we proposed `liberty -> dev`

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
turkeylurkey: It would cause the loose war config file to be generated using the project variable io.openliberty.tools.projectRoot which could be put to some productive use but it is not really meant to be used this way. In particular you would have to define the variable (in bootstrap.properties, server.xml or command line) manually to run the app in the server. 
Another way to pass the information would be a Java System property. Or a temp file. What makes the most sense in a Gradle environment? Are there other situations that require an option value in multiple tasks?

cherylking: @mattbsox Can you chime in here? I don't thinking having this container option exposed on the DeployTask makes sense.

src/main/groovy/io/openliberty/tools/gradle/extensions/DevExtension.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.openliberty.tools.gradle.extensions
+
+class DevExtension {
+    boolean container = false
+    String dockerfile
```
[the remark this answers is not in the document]
ericglau: DevTask has the `dockerfile` option with parameter type String.  According to https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/options/Option.html, you cannot have a File type option.  For consistency, the DevExtension uses String as well.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
mattbsox: I think passing the value as a project property on the command line would work.

Should be able to set it like this: `-Pliberty.dev.container=true`

All the tasks have access to the Liberty extension so it will be shared across them. You can use the same line in the `runGradleTask()` parameters as well.

src/main/groovy/io/openliberty/tools/gradle/extensions/DevExtension.groovy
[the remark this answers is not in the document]
ericglau: We will change the DevExtension to use File.

src/main/groovy/io/openliberty/tools/gradle/LibertyTaskFactory.groovy
```diff
@@ -36,6 +36,7 @@ import io.openliberty.tools.gradle.tasks.CleanTask
 import io.openliberty.tools.gradle.tasks.CompileJSPTask
 import io.openliberty.tools.gradle.tasks.arquillian.ConfigureArquillianTask
 import io.openliberty.tools.gradle.tasks.DevTask
+import io.openliberty.tools.gradle.tasks.DevcTask
```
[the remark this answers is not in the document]
ericglau: done

src/main/groovy/io/openliberty/tools/gradle/extensions/DevExtension.groovy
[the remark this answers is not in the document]
ericglau: done

src/main/groovy/io/openliberty/tools/gradle/extensions/LibertyExtension.groovy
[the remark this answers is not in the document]
ericglau: Updated copyright.  Left the extension as `liberty -> dev` as reasoned above.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
ericglau: Changed to use buildDir.  For cases where buildDir is outside of the project root, that will be handled separately in https://github.com/OpenLiberty/ci.maven/issues/876 , but for now it will just use `build/libs` if that happens since it must be within the project root that is mounted.

cherylking: Don't you need to open a gradle issue also to handle buildDir outside project root?

ericglau: Opened https://github.com/OpenLiberty/ci.gradle/issues/489

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
ericglau: We've removed the extra booleans and changed it to use `project.liberty.dev.container` as suggested in Matt's comment below.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
ericglau: This has been changed as suggested above to use a project property and access the Liberty extension.

ericglau:
@cherylking All of the previous review comments have been addressed.  Please re-review with the latest changes, thanks.

docs/libertyExtensions.md
```diff
@@ -51,3 +54,15 @@ The following properties are supported for server configuration.
 | timeout | String | 1.0 | Waiting time before the server starts. The default value is 30 seconds. The unit is seconds. Only used with `libertyStart` and `deploy` tasks. | No |
 | var | Properties | 3.0 | Inline server variables that are written to the `configDropins/overrides/liberty-plugin-variable-config.xml` file in the server directory. The property name is used for the variable `name`, and the property value is used for the variable `value`.| No|
 | verifyAppStartTimeout | int | 2.0 | Wait time for checking message logs for start of all applications installed with the `deploy` task. Only used with the `libertyStart` task. Default value is 0 seconds with no verification. | No |
+
```
cherylking, line 57: I don't think we need to document these here again. You already documented them in the dev/devc task document. Just add something to line 8 above where it mentions compileJSP and installLiberty can be configured with their related extension in side the `liberty` block. Then remove the changes to lines 3 and 6 also.

docs/libertyDev.md
```diff
@@ -80,3 +82,58 @@ test {
...
+        dockerfile = file('myDockerfile')
+    }
+}
+```
+
+### Command Line Parameters
+
+These parameters are available in addition to the ones in the `libertyDev` section above.
+
+| Parameter | Description | Required |
```
cherylking, line 135: Use the following header/columns like all the other tasks:

| Attribute | Type | Since | Description | Required |

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -26,6 +26,7 @@ import org.apache.commons.io.FilenameUtils
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.tasks.options.Option
```
cherylking, line 29: Is this needed?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -158,6 +161,39 @@ class DevTask extends AbstractServerTask {
...
+    }
+
+    private File dockerfile;
+
+    @Option(option = 'dockerfile', description = 'Dev mode will build a docker image from the provided Dockerfile and start a container from the new image.')
+    void setDockerfile(String dockerfile) {
+        if (dockerfile != null) {
+            File temp = new File(dockerfile);
+            try {
+                this.dockerfile = new File(temp.getCanonicalPath()); // ensures the dockerfile is defined with the full path - matches how maven behaves
```
cherylking, line 183: I thought setting a dockerfile or dockerRunOpts was supposed to automatically set container to true, but I don't see any code doing that.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
ericglau: Removed

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
ericglau: Fixed by adding a `processContainerParams()` method

docs/libertyDev.md
ericglau: Fixed

docs/libertyExtensions.md
ericglau: Done

docs/libertyExtensions.md
```diff
@@ -1,16 +1,13 @@
...
-These properties are divided in three groups:
+These properties are divided in two groups:
 * The [general runtime properties](#general-runtime-properties) which control how the runtime is installed with the `installLiberty` task.
 * The [Liberty server configuration](#liberty-server-configuration) properties control how a server is initialized and the applications that are installed. These properties are defined in a `server` block within the `liberty` block.
-* The [dev extension properties](#dev-extension-properties) that are specific to dev mode when using containers. These properties are defined in a `dev` block within the `liberty` block.
 
-The [installLiberty](installLiberty.md) and [compileJSP](compileJsp.md) tasks can be configured with their related extension inside the `liberty` block.
+The [installLiberty](installLiberty.md) and [compileJSP](compileJsp.md) tasks can be configured with their related extension inside the `liberty` block.  The [libertyDevc](libertyDev.md#libertydevc-task-container-mode) task can be configured with its related extension inside the `dev` block.
```
cherylking, line 7: Combine these sentences into one.
The [installLiberty](installLiberty.md), [compileJSP](compileJsp.md) and [libertyDevc](libertyDev.md#libertydevc-task-container-mode) tasks can be configured with their related extension inside the `liberty` block.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -681,6 +741,29 @@ class DevTask extends AbstractServerTask {
...
+                setContainer(DEFAULT_CONTAINER);
+            } else {
+                setContainer(buildContainerSetting);
+            }
+        }
+
+        if (dockerfile == null) {
+            File buildDockerfileSetting = project.liberty.dev.dockerfile; // get from build.gradle
+            if (buildDockerfileSetting != null) {
+                setDockerfile(buildDockerfileSetting.getAbsolutePath()); // setDockerfile will convert it to canonical path
```
cherylking, line 757: Should this set container to true?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -681,6 +741,29 @@ class DevTask extends AbstractServerTask {
...
+            }
+        }
+
+        if (dockerRunOpts == null) {
+            String buildDockerRunOptsSetting = project.liberty.dev.dockerRunOpts; // get from build.gradle
+            if (buildDockerRunOptsSetting != null) {
+                setDockerRunOpts(buildDockerRunOptsSetting);
```
cherylking, line 764: Should this set container to true? I'm thinking the setDockerfile and setDockerRunOpts methods should just call setContainer(true). Would that make sense?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -792,6 +882,25 @@ class DevTask extends AbstractServerTask {
         }
     }
 
+    private void processContainerParams() throws Exception {
```
cherylking, line 885: If setDockerfile and setDockerRunOpts called setContainer(true), this would not be necessary right? Should still check that dockerfile.exists() though.

ericglau: Refactored to put processContainerParams into initializeDefaultValues, and moved the other container related param processing into there as well. Commit: https://github.com/OpenLiberty/ci.gradle/pull/477/commits/e56e331c5e9569bae1cc2daeac79daba72cc5b08

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
ericglau: This only handles when the dockerRunOpts or dockerfile params are not set in command line but set in the dev extension.  We also need to address when the params are set in the command line.  The logic in processContainerParams handles both cases.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
ericglau: See other comment

docs/libertyExtensions.md
ericglau: Fixed

Issue #473, opened 2020-06-22
Set copyLibsDirectory when deploying loose app with container support


ericglau:
Loose application with container support currently parameterizes the project root, but dependencies that are outside of the project dir still use absolute paths. To solve this, set the copyLibsDirectory when calling the deploy task with container option, so that the dependencies will be copied to somewhere under the project root.

Gradle equivalent of https://github.com/OpenLiberty/ci.maven/issues/834

Depends on https://github.com/OpenLiberty/ci.gradle/issues/468

ericglau:
Fixed

PR #481, opened 2020-06-26
In container mode, ensure copyLibsDirectory option is set to default or user value.

turkeylurkey:
#473

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
@@ -293,23 +293,25 @@ class DeployTask extends AbstractServerTask {
         }
 
         if (container) {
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -294,6 +295,22 @@ class DeployTask extends AbstractServerTask {
         if (container) {
             config.setProjectRoot(task.getProject().getProjectDir().getAbsolutePath());
```
ericglau, line 296: Although this line wasn't changed from this PR, use getCanonicalPath here to be consistent with the below, similar to https://github.com/OpenLiberty/ci.maven/blob/master/liberty-maven-plugin/src/main/java/io/openliberty/tools/maven/applications/DeployMojoSupport.java#L118

turkeylurkey: I agree.

PR #485, opened 2020-07-01
Improve installDir with warnings and errors

### LINKED ISSUE #432: Give error message if installDir location is not valid
If `installDir` is specified and no Liberty server is at that location, the installLiberty task tries to install the default OL kernel instead with no indication to the user that there is a problem. This behavior seems incorrect.

We should detect that a valid Liberty server was not found at the specified location, and give an appropriate error message.

kevinwlau:
Fixes #432

- Update documentation to specify using `wlp` in the `installDir` property
- Add warning message when the specified `installDir` does not contain `wlp` in the path
  - and automatically checks the specified path with `wlp` appended.
- Throw error when `installDir` does not correctly resolve to a valid Liberty installation
- Add test to confirm error shows when `installDir` points to invalid Liberty installation

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
@@ -178,7 +178,7 @@ class Liberty implements Plugin<Project> {
         if (installDir.endsWith("wlp")) {
             return installDir
         } else { // not valid wlp dir
-            project.getLogger().warn(MessageFormat.format("The installDir {0} path does not reference a wlp folder. Checking for a valid Liberty installation at path {0}/wlp instead.", installDir))
+            project.getLogger().warn(MessageFormat.format("The installDir {0} path does not reference a wlp folder. Using path {0}/wlp instead.", installDir))
             return new File(installDir, 'wlp').toString()
         }
     }
--- a/src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
@@ -29,6 +29,7 @@ import org.gradle.api.GradleException
 import org.gradle.api.tasks.testing.Test
 
 import java.util.Properties
+import java.text.MessageFormat
 
 class Liberty implements Plugin<Project> {
 
@@ -169,6 +170,7 @@ class Liberty implements Plugin<Project> {
             if (project.liberty.installDir.endsWith("wlp")) {
                 installDir = new File(project.liberty.installDir)
             } else { // not valid wlp dir
+                project.getLogger().warn(MessageFormat.format("The installDir {0} is not pointing to a wlp folder. Trying {0}/wlp", project.liberty.installDir))
                 installDir = new File(project.liberty.installDir, 'wlp')
             }
 
--- a/src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corporation 2014, 2018.
+ * (C) Copyright IBM Corporation 2014, 2020.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
@@ -166,19 +166,20 @@ class Liberty implements Plugin<Project> {
                return new File(project.liberty.baseDir, 'wlp')
            }
         } else { // installDir is specified
-            File installDir
-            if (project.liberty.installDir.endsWith("wlp")) {
-                installDir = new File(project.liberty.installDir)
-            } else { // not valid wlp dir
-                project.getLogger().warn(MessageFormat.format("The installDir {0} path does not reference a wlp folder. Checking for a valid Liberty installation at path {0}/wlp instead.", project.liberty.installDir))
-                installDir = new File(project.liberty.installDir, 'wlp')
-            }
+            String installDirPath = checkAndAppendWlp(project)
+            File installDir = new File(installDirPath)
 
-            //check if installDir is a valid Liberty installation
-            if (!(new File(installDir, "lib/ws-launch.jar").exists())) {
-                throw new GradleException("Unable to find a valid Liberty installation at installDir path " + project.liberty.installDir +". Please specify a valid path for the installDir property.")
-            }
             return installDir;
+        }
+    }
+
+    private static String checkAndAppendWlp(Project project) {
+        String installDir = project.liberty.installDir
+        if (installDir.endsWith("wlp")) {
+            return installDir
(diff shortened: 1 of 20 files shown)


src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
```diff
@@ -164,8 +165,19 @@ class Liberty implements Plugin<Project> {
...
                return new File(project.liberty.baseDir, 'wlp')
            }
-        } else {
-           return new File(project.liberty.installDir)
+        } else { // installDir is specified
+            File installDir
+            if (project.liberty.installDir.endsWith("wlp")) {
+                installDir = new File(project.liberty.installDir)
+            } else { // not valid wlp dir
+                project.getLogger().warn(MessageFormat.format("The installDir {0} is not pointing to a wlp folder. Trying {0}/wlp", project.liberty.installDir))
```
cherylking, line 173: The installDir {0} path does not reference a wlp folder. Checking for a valid Liberty installation at path {0}/wlp instead.

src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
```diff
@@ -164,8 +165,19 @@ class Liberty implements Plugin<Project> {
...
+                installDir = new File(project.liberty.installDir, 'wlp')
+            }
+
+            //check if installDir is a valid Liberty installation
+            if (!(new File(installDir, "lib/ws-launch.jar").exists())) {
+                throw new GradleException("Unable to find valid Liberty installation at the specified path. Please specify a valid installDir.")
```
cherylking, line 179: "Unable to find a valid Liberty installation at installDir path " + project.liberty.installDir +". Please specify a valid path for the installDir property."

kevinwlau:
Thanks for those catches and message updates.

> Are you going to add a test where installDir points to a valid Liberty installation without wlp in the path?

Yes, working on that now, and I may have found that the current logic is flawed.

`getInstallDir` seems like the proper place to check specifically the user-specified `installDir`, **but** this also gets called during project configuration before running any project dependencies. 
This leads to throwing the error while configuring the project before it gets a chance to run the dependencies.

Solution to above is to move the error check to run only on the `installLiberty` task. However, by doing that, a default `installDir` is set even when it is not specified, and the check ends up checking the default `installDir` instead of looking for a user-configured one. Obviously Liberty is not installed yet and it will error.

Looking into having a different method entirely, specifically for the `installLiberty` task, to avoid overlap with the existing logic that is used in other tasks and project configuration.

cherylking:
I have to say I never liked that the check/error were in the same method that was getting the installation directory. If you need to talk through the rewrite, let me know.

kevinwlau:
Currently everything should work correctly for the `installLiberty` task.

Will have other tasks use the new Liberty install validation after further testing.

src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
```diff
@@ -164,8 +165,21 @@ class Liberty implements Plugin<Project> {
...
+            return installDir;
+        }
+    }
+
+    private static String checkAndAppendWlp(Project project) {
+        String installDir = project.liberty.installDir
+        if (installDir.endsWith("wlp")) {
+            return installDir
+        } else { // not valid wlp dir
+            project.getLogger().warn(MessageFormat.format("The installDir {0} path does not reference a wlp folder. Checking for a valid Liberty installation at path {0}/wlp instead.", installDir))
```
cherylking, line 181: Let's change this warning message since we are no longer doing the validation checking for the installation here.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractTask.groovy
```diff
@@ -107,6 +100,22 @@ abstract class AbstractTask extends DefaultTask {
         return (installDir.exists() && new File(installDir, "lib/ws-launch.jar").exists())
     }
 
+    @Internal
+    protected boolean isLibertyInstalledAndValid(Project project) {
+        // only used for installLibertyTask, where it should throw an exception instead of returning false when the installDir is invalid.
+        if(project.liberty.installDir != null) {
+            project.getLogger().info("Found user-specified installDir. Checking if valid " + project.liberty.installDir)
```
cherylking, line 107: I think this log line should be removed. The call to checkAndAppendWlp method will potentially give a warning, and if no Liberty installation is found in this method we send an error.

src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
kevinwlau: How about just removing the word "valid"?

cherylking: It's not checking the installation here. Change the second sentence to "Using path {0}/wlp instead."

src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
```diff
@@ -29,6 +29,7 @@ import org.gradle.api.GradleException
 import org.gradle.api.tasks.testing.Test
 
 import java.util.Properties
+import java.text.MessageFormat
```
cherylking, line 32: Update copyright year.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractTask.groovy
```diff
@@ -102,9 +95,18 @@ abstract class AbstractTask extends DefaultTask {
         return task
     }
     @Internal
-    protected boolean isLibertyInstalled(Project project) {
-        File installDir = getInstallDir(project)
-        return (installDir.exists() && new File(installDir, "lib/ws-launch.jar").exists())
+    protected boolean isLibertyInstalledAndValid(Project project) {
+        if(project.liberty.installDir != null) {
+            String installDirPath = Liberty.checkAndAppendWlp(project)
```
cherylking, line 100: Why are we not using getInstallDir(project) here too?
```
File installDir = getInstallDir(project)
boolean installationExists = installDir.exists() && new File(installDir,"lib/ws-launch.jar").exists()
if (!installationExists && (project.liberty.installDir != null)) {
    throw new GradleException(...)
}
return installationExists
```

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -251,7 +251,7 @@ class DevTask extends AbstractServerTask {
 
         @Override
         public void stopServer() {
-            if (isLibertyInstalled(project)) {
+            if (isLibertyInstalledAndValid(project)) {
```
cherylking, line 254: update copyright year

src/main/groovy/io/openliberty/tools/gradle/tasks/StopTask.groovy
```diff
@@ -30,7 +30,7 @@ class StopTask extends AbstractServerTask {
 
     @TaskAction
     void stop() {
-        if (isLibertyInstalled(project)) {
+        if (isLibertyInstalledAndValid(project)) {
```
cherylking, line 33: update copyright year

PR #486, opened 2020-07-02
Generated server.env can now be merged with configured server.env or inline properties

### LINKED ISSUE #459: Generated keystore_password in server.env overwritten when combined with env configuration
It looks like the logic in `AbstractServerTask.copyConfigFiles` is overwriting the generated values stored in the `server.env` file during server creation. If a `server.env` exists in the default config directory, or if the server extension has inline values or a file, the generated `keystore_password` value is always lost.

jjvilleg:
Resolves #459 . 
Working on tests


src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -285,17 +285,45 @@ abstract class AbstractServerTask extends AbstractTask {
 
         // envProjectProps and serverEnvFile take precedence over server.env from configDirectory
         File envFile = new File(serverDirectory, "server.env")
+        Properties serverEnvProps = convertServerEnvToProperties(envFile);
```
cherylking, line 288: You need code at line 240 to check for appendingServerEnv. That code is copying over the server.env located in the default configuration directory (src/main/liberty/config) and it will overwrite any server.env file already in the target server directory.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -731,7 +787,32 @@ abstract class AbstractServerTask extends AbstractTask {
         }
     }
 
-    private void writeServerEnvProperties(File file, Properties properties, Properties projectProperties) throws IOException {
+    private Properties convertServerEnvToProperties(File serverEnv) {
+        if(serverEnv == null) {
+            return null;
```
cherylking, line 792: I think this is a bad idea to return null. Also, perhaps you should check that `if ((serverEnv == null) || !serverEnv.exists())`. I would return the empty Properties object that you are creating at line 796 in this situation.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -285,17 +313,45 @@ abstract class AbstractServerTask extends AbstractTask {
 
         // envProjectProps and serverEnvFile take precedence over server.env from configDirectory
         File envFile = new File(serverDirectory, "server.env")
+        Properties serverEnvProps = convertServerEnvToProperties(envFile);
+        
         if ((server.env != null && !server.env.isEmpty()) || !envProjectProps.isEmpty()) {
-            if (serverEnvPath != null) {
-                logger.warn("The " + serverEnvPath + " file is overwritten by inlined configuration.")
+            Properties configuredProps = combineServerEnvProperties(server.env, envProjectProps);
```
cherylking, line 319: So I wasn't sure where to talk about this, but with the way this code is written it appends server env properties from multiple sources except certain combinations....which I think will be confusing for the customer. I'm going to try to capture this in this comment, but may need to create a table outside of here and attach an image later.

The first section of code dealing with the config directory will append the server.env from configDirectory with the default server.env. Then this code here will append the inline server env variables (from server.env Properties or project properties) with the potentially already appended server.env on the target server. If they also specified a serverEnvFile, that will not get appended because we said the inline configuration takes precedence. But in this appending scenario I think that is confusing.

Finally, if they did not specify inline configuration (through server.env Properties or project Properties) but did specify serverEnvFile, then that will get appended to the server.env in the target server.

I think we either need to append all configured server.env properties (no matter where they came from), or only the configured ones that previously took precedence. The previous precedence was (with 1 being the highest precedence):

1. project properties
2. server.env properties
3. serverEnvFile
4. server.env from configDirectory

cherylking:
Screen Shot 2020-07-08 at 9 30 48 AM

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -746,6 +827,10 @@ abstract class AbstractServerTask extends AbstractTask {
             combinedEnvProperties = properties
         }
 
+        return combinedEnvProperties;
```
cherylking, line 830: One more comment...the way this combined properties before was fine because it just used them locally to write the file. But you are potentially taking the result and calling combineServerEnvProperties again with it. So setting the combinedEnvProperties to either projectProperties or properties can have some interesting results later (with the original Properties objects getting altered). So let's create a new Properties object (at line 816) and do a putAll at lines 819 and 827 to be safe.

src/main/groovy/io/openliberty/tools/gradle/extensions/ServerExtension.groovy
```diff
@@ -34,6 +34,7 @@ class ServerExtension {
     File bootstrapPropertiesFile
     File jvmOptionsFile
     File serverEnvFile
+    boolean appendServerEnv = false;
```
cherylking, line 37: Update copyright year.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -200,6 +201,32 @@ abstract class AbstractServerTask extends AbstractTask {
         }
     }
 
+    protected void copyConfigDirectory() {
+        //merge default server.env with one in config directory
+        if(server.configDirectory != null && server.appendServerEnv) {
```
cherylking, line 206: Not sure why you are checking if `server.configDirectory != null` here. The code checked that `server.configDirectory.exists()` before calling this method. Also, I think you should do line 210 before this if statement and change this if statement to the following:

`if (configDirServerEnv.exists() && server.appendServerEnv)`

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -284,20 +312,9 @@ abstract class AbstractServerTask extends AbstractTask {
...
-            writeServerEnvProperties(envFile, server.env, envProjectProps)
-            serverEnvPath = "inlined configuration"
-        } else if (server.serverEnvFile != null && server.serverEnvFile.exists()) {
-             if (serverEnvPath != null) {
-                logger.warn("The " + serverEnvPath + " file is overwritten by the " + server.serverEnvFile.getCanonicalPath() + " file.")
-            }
-            Files.copy(server.serverEnvFile.toPath(), envFile.toPath(), StandardCopyOption.REPLACE_EXISTING)
-            serverEnvPath = server.serverEnvFile.getCanonicalPath()
-        }
+        setServerEnv(envFile, serverEnvPath)
```
cherylking, line 317: I would call this `handleServerEnvFileAndProperties` since it is not actually setting something like a set method typically would. Also, it doesn't seem necessary to do line 316 here.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -731,8 +748,91 @@ abstract class AbstractServerTask extends AbstractTask {
...
+                serverEnvProps.put(key,value);
+            }
+        }
+
+        return serverEnvProps;
+    }
+
+    private Properties combineServerEnvProperties(Properties properties, Properties projectProperties) {
         Properties combinedEnvProperties = null
+        //Create new properties object
```
cherylking, line 835: Please change line 834 to create a new Properties object. Then change line 838 to do a `combinedEnvProperties.putAll(projectProperties)` and line 846 to do a `combinedEnvProperties.putAll(properties)`. Then you can remove line 840.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -731,8 +748,91 @@ abstract class AbstractServerTask extends AbstractTask {
...
+                logger.warn("The " + serverEnvPath + " file is merged with inlined configuration.")
+            }
+
+            //Create properties from existing server env
+            //Merge server.env props with generated serverEnvPath
+            mergedProperties = combineServerEnvProperties(serverEnvProps, configuredProps);
+            // writeServerEnvProperties(envFile, mergedProperties);
+        }
+
+        if (server.serverEnvFile != null && server.serverEnvFile.exists()) {
```
cherylking, line 777: This is merging the serverEnvFile last which is not correct. The project properties and server.env take precedence. You need to do this in the following order.

1. Combine the serverEnvFile properties with the envFile properties (serverEnvProps in your method).
2. Combine the configuredProps (which are the envProjectProps combined with the server.env properties) with the combined properties from step 1.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
jjvilleg: Had missed this from last review. Should be in the next commit 👍

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -763,26 +761,21 @@ abstract class AbstractServerTask extends AbstractTask {
...
                 logger.warn("The " + serverEnvPath + " file is merged with the " + server.serverEnvFile.getCanonicalPath() + " file.")
             }
             Properties configuredServerEnvProps = convertServerEnvToProperties(server.serverEnvFile);
             //Merge with either default server.env or with what has already been merged if
-            Properties propsToMergeWith = mergedProperties.isEmpty() ? serverEnvProps : mergedProperties;
+            mergedProperties = (Properties) combineServerEnvProperties(serverEnvProps, configuredServerEnvProps);
+        }
 
-            mergedProperties = (Properties) combineServerEnvProperties(propsToMergeWith, configuredServerEnvProps);
+        if ((server.env != null && !server.env.isEmpty()) || !envProjectProps.isEmpty()) {
```
cherylking, line 773: I think this can be simplified to just check `if (!configuredProps.isEmpty())`

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -831,19 +824,18 @@ abstract class AbstractServerTask extends AbstractTask {
     }
 
     private Properties combineServerEnvProperties(Properties properties, Properties projectProperties) {
-        Properties combinedEnvProperties = null
-        //Create new properties object
+        Properties combinedEnvProperties = new Properties()
         if (! projectProperties.isEmpty()) {
             if (properties.isEmpty()) {
-                combinedEnvProperties = projectProperties
+                combinedEnvProperties.putAll(projectProperties)
```
cherylking, line 830: Remove the initialization at line 832 in the else statement.

src/test/resources/sample.servlet/testAppendServerEnvWithEnvProps.gradle
```diff
@@ -0,0 +1,94 @@
...
+        serverXmlFile = file("src/main/liberty/config/server-apps-test.xml")
+        name = wlpServerName
+        deploy {
+            apps = [war]
+            copyLibsDirectory = file("${project.buildDir}/libs")
+        }
+        appendServerEnv = true
+        serverEnvFile = file("src/resources/server.env")
+        configDirectory = file('src/resources/config')
+        env = ['TEST_PROP_2':'white', 'CONFIG_SERVER_ENV_PROPS':'TEST']
```
cherylking, line 53: To prove that the env props take precedence over the serverEnvFile, I would like the same property to appear in both with a different value. The test case as it stands proves that all four sources got merged, and that env props take precedence over configDir and serverEnvFile takes precedence over configDir.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -200,6 +201,31 @@ abstract class AbstractServerTask extends AbstractTask {
...
     }
 
+    protected void copyConfigDirectory() {
+        //merge default server.env with one in config directory
+        File configDirServerEnv = new File(server.configDirectory, "server.env")
+        if (configDirServerEnv.exists() && server.appendServerEnv) {
+            FileFilter fileFilter =   FileFilterUtils.notFileFilter(FileFilterUtils.nameFileFilter("server.env", null))
+            FileUtils.copyDirectory(server.configDirectory, getServerDir(project), fileFilter)
+
+            if(configDirServerEnv.exists()) {
```
cherylking, line 211: Just noticed that you don't need to check this again. It was already checked at line 207.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -731,21 +745,103 @@ abstract class AbstractServerTask extends AbstractTask {
...
+        }
+    }
+
+    private void setServerEnvHelper(File envFile, String serverEnvPath, Properties configuredProps) {
+        if ((server.env != null && !server.env.isEmpty()) || !envProjectProps.isEmpty()) {
+            if (serverEnvPath != null) {
+                logger.warn("The " + serverEnvPath + " file is overwritten by inlined configuration.")
+            }
+            writeServerEnvProperties(envFile, configuredProps)
+            serverEnvPath = "inlined configuration"
```
cherylking, line 792: Strings are immutable. Setting this has no affect on the serverEnvPath variable that was passed in. Perhaps we need to be passing this back as the return value?

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -731,21 +745,103 @@ abstract class AbstractServerTask extends AbstractTask {
...
+        }
+    }
+
+    private void setServerEnvWithAppendServerEnvHelper(File envFile, String serverEnvPath, Properties configuredProps) {
+        Properties serverEnvProps = convertServerEnvToProperties(envFile);
+        Properties mergedProperties = new Properties();
+
+        if (server.serverEnvFile != null && server.serverEnvFile.exists()) {
+            if (serverEnvPath != null) {
+                logger.warn("The " + serverEnvPath + " file is merged with the " + server.serverEnvFile.getCanonicalPath() + " file.")
```
cherylking, line 766: Need to set and return serverEnvPath to indicate what happened in this method. I think this would be more accurate to say "The serverEnvFile "+ server.serverEnvFile.getCanonicalPath()+" is merged with the "+serverEnvPath+" file.".

cherylking: Also, even when serverEnvPath is null, we are still merging with the default server.env. So a log message would be good in that case too.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -731,21 +745,103 @@ abstract class AbstractServerTask extends AbstractTask {
...
+            }
+            Properties configuredServerEnvProps = convertServerEnvToProperties(server.serverEnvFile);
+            //Merge with either default server.env or with what has already been merged if
+            mergedProperties = (Properties) combineServerEnvProperties(serverEnvProps, configuredServerEnvProps);
+        }
+
+        if ((server.env != null && !server.env.isEmpty()) || !envProjectProps.isEmpty()) {
+            if (serverEnvPath != null) {
+                logger.warn("The " + serverEnvPath + " file is merged with inlined configuration.")
+            }
```
cherylking, line 776: Need to set and return serverEnvPath at the end of this method to indicate what happened in this method. Seeing how this method could do both things (merge serverEnvFile with either the configDir server.env or the default server.env AND merge env props with those as well), I think we should set a flag in each block indicating what got merged and then decide at the end of this method what to set serverEnvPath to. It could be one of the following:

The log message that uses serverEnvPath is as follows:             
logger.info("Update server configuration file server.env from " + serverEnvPath)

1) merging env properties, serverEnvFile server.serverEnvFile.getCanonicalPath(), and configDir server.env serverEnvPath (this is the original one passed in if not null)
2) merging env properties and configDir server.env serverEnvPath (this is the original one passed in if not null)
3) merging serverEnvFile server.serverEnvFile.getCanonicalPath() and configDir server.env serverEnvPath (this is the original one passed in if not null)

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
jjvilleg: Will do. Especially since I'm modifying `setServerEnvWithAppendServerEnvHelper` to also return this value.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
jjvilleg: Should I remove the previous logger.warn messages, or keep them in addition to a logger.info message you suggested?

cherylking: Maybe they should be logger.debug for additional info if they are confused with the result. The logger.info at the end of the main method should be sufficient most of the time.

jjvilleg: As I was writing this, I realized there are a few more cases to account for. 
4. merging env properties and default server.env and serverEnvFile server.serverEnvFile.getCanonicalPath() (without config dir)

And in all cases, they are also being merged with the default server.env

cherylking: Yes, but "Update server configuration file server.env" is the default server.env.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -745,57 +742,85 @@ abstract class AbstractServerTask extends AbstractTask {
...
         if(!mergedProperties.isEmpty()) {
             writeServerEnvProperties(envFile, mergedProperties);
+            return setServerEnvPathHelperForAppendServerEnv(envFile, configuredProps, serverEnvPath)
+        }
+
+        return serverEnvPath;
+    }
+
+    private String setServerEnvPathHelperForAppendServerEnv(File envFile, Properties configuredProps, String serverEnvPath) {
+        boolean configDirEnvMerged = envFile != null;
```
cherylking, line 793: envFile is the server.env in the server directory, which may or may not have had a server.env from the configDirectory merged with it.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -745,57 +742,85 @@ abstract class AbstractServerTask extends AbstractTask {
...
+            return "merging env properties, serverEnvFile" +  server.serverEnvFile.getCanonicalPath() + ", and configDir server.env" +  serverEnvPath
+        }
+        if (configDirEnvMerged && serverEnvFileMerged) {
+            return "merging serverEnvFile" +  server.serverEnvFile.getCanonicalPath() + ", and configDir server.env" +  serverEnvPath
+        }
+        if (configDirEnvMerged && inlineEnvPropsMerged) {
+            return "merging env properties and configDir server.env" + serverEnvPath 
+        }
+        if (serverEnvFileMerged && inlineEnvPropsMerged) {
+            return "merging env properties and serverEnvFile" +  server.serverEnvFile.getCanonicalPath()
```
cherylking, line 807: There's actually more combinations than this I think. Perhaps we can use an approach to append to the String that gets returned. Start with "merging" and then for each boolean you check add " abc," then " cde," etc.
Once you've checked each boolean and added that part to the String, replace the last `,` with a `.` and return the String.

jjvilleg: Yeah, that's what I was thinking too. Can use a StringBuilder for this 👍

docs/libertyExtensions.md
```diff
@@ -41,6 +41,7 @@ The following properties are supported for server configuration.
 | jvmOptions| List | 2.0 | Inline `List` of jvm options that is written to the jvm.options file in the server directory. These properties take precedence over a specified jvm.options file.| No|
 | jvmOptionsFile| File | 2.0 | Location of the file containing JVM options to copy to the jvm.options file in the server instance.| No|
 | looseApplication | boolean | 2.0 | Indicates to install application using loose application configuration so that war or ear files do not need to be regenerated for every application update. The loose application support for ear files requires Gradle 4.0 or higher.  The default is `true`. | No |
+| mergeServerEnv | boolean | 3.0 | Merge server.env properties from a server.env specified in the `configDirectory`, server.env file specified by the `serverEnvFile` property, or from in line properties specified in the `env` attribute with the generated server.env. Conflicts are resolved with the same precedence as the replacement policy when this property is `false`. A server.env in the `configDirectory` is of lowest precedence, a `serverEnvFile` being the next highest, and `env` inline being of highest precedence.  `false`. | No |
```
cherylking, line 44: Change the `Since` version to 3.1-M1. We will change it to 3.1 once that is released.

Suggested rewording: Merge the server environment properties from all specified sources with the default generated `server.env` file in the target server. Conflicts are resolved with the same precedence as the replacement policy when this attribute is set to `false`. The properties specified in the `env` attribute are highest precedence, followed by the `serverEnvFile` attribute, then the `server.env` file located in the `configDirectory`, and finally the default generated `server.env` file in the target server. The default value is `false`.

Issue #497, opened 2020-07-15
Appveyor tests are failing


evie-lau:
It seems the main issues for the appveyor tests are:

- tests that check for `/commons-text-1.1.jar` in the TestLoose*Application tests
- PollingDevTest as mentioned in #439

cherylking:
We previously moved the PollingDevTest to the Gradle 4.10 wrapper section of the build as they were passing there. But now they fail due to an API that is not available on that level. I'm going to remove the PollingDevTest from the AppVeyor builds because of this.

<img width="913" alt="Screen Shot 2020-11-19 at 8 04 55 AM" src="https://user-images.githubusercontent.com/29436451/99676510-37f3de00-2a3e-11eb-999d-826e0a40f149.png">

cherylking:
We no longer run tests on Appveyor.

PR #498, opened 2020-07-16
Exclude failing tests on appveyor

kevinwlau:
Fix for #497 until we investigate the root cause

Changes:
- Move PollingDevTest to use gradle 4.10
- Exclude TestLoose* tests (3 of the 4 fail on checking for `commons-text-1.1.jar`

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,7 +35,7 @@ before_build:
 build_script:
     - cmd: |
         set GRADLE_OPTS="-Dorg.gradle.jvmargs='-XX:MaxMetaspaceSize=512m'"
-        gradlew.bat clean install check -Ptest.exclude="**/*15*" -Druntime=%RUNTIME% -DruntimeVersion=%RUNTIME_VERSION% --stacktrace --info --no-daemon
+        gradlew.bat clean install check -Ptest.exclude="**/*15*,**/Polling*,**/*estLoose*" -Druntime=%RUNTIME% -DruntimeVersion=%RUNTIME_VERSION% --stacktrace --info --no-daemon
         gradlew.bat wrapper --gradle-version 4.10
         gradlew.bat check -Ptest.include="**/*15*" -Druntime=%RUNTIME% -DruntimeVersion=%RUNTIME_VERSION% --stacktrace --info --no-daemon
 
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -35,7 +35,7 @@ before_build:
 build_script:
     - cmd: |
         set GRADLE_OPTS="-Dorg.gradle.jvmargs='-XX:MaxMetaspaceSize=512m'"
-        gradlew.bat clean install check -Ptest.exclude="**/*15*" -Druntime=%RUNTIME% -DruntimeVersion=%RUNTIME_VERSION% --stacktrace --info --no-daemon
+        gradlew.bat clean install check -Ptest.exclude="**/*15*,**/Polling*,**/*estLoose*" -Druntime=%RUNTIME% -DruntimeVersion=%RUNTIME_VERSION% --stacktrace --info --no-daemon
```
cherylking, line 38: Curious why not **/TestLoose*? Did you leave the T off on purpose? Also, can you open an issue specifically listing the tests that this PR excluded and need to be investigated (and reference this PR in that issue)?

appveyor.yml
```diff
@@ -35,7 +35,7 @@ before_build:
...
         gradlew.bat wrapper --gradle-version 4.10
         gradlew.bat check -Ptest.include="**/*15*" -Druntime=%RUNTIME% -DruntimeVersion=%RUNTIME_VERSION% --stacktrace --info --no-daemon
```
cherylking, line 40: One more thought...can you see if any of the tests you are excluding actually work with this older gradle wrapper?

appveyor.yml
kevinwlau: For some reason, I seemed to have problems in the past getting it to filter correctly without the first wildcard. I can try again to make it cleaner.
#497 already lists the issues and shouldn't be auto-closed by this PR.

appveyor.yml
kevinwlau: Seems like the TestLoose* still fails when using gradle 4.10, and the PollingDevTest ran fine until it errored on cleanup.

Will see if any of the other jobs or reruns complete with a success on PollingDevTest.

PR #503, opened 2020-07-22
Updated project root and added code to generate dev mode config

TrevCraw:


ericglau:
One of the travis builds passed, but the rest failed because ci.common 1.8.7 was released so ci.common master is no longer at 1.8.7-SNAPSHOT.  We should update ci.gradle to use ci.common 1.8.7 (released version).

cherylking:
I've released ci.common 1.8.7 but it is not in Maven Central yet. It sometimes takes a few hours.

Issue #479, opened 2020-06-25
Document the SpringBootExtension


cherylking:
Support was added for SpringBoot Uber jar applications in 2.6 release. Here is the extension for it: https://github.com/OpenLiberty/ci.gradle/blob/master/src/main/groovy/io/openliberty/tools/gradle/extensions/SpringBootExtension.groovy

Need to create a document describing that extension and add a link to it either at the bottom of the readme, in the `deploy.md`, or in the `libertyExtensions.md`.

mattbsox:
Docs for our Maven plugin SpringBoot support: https://github.com/OpenLiberty/ci.maven/blob/master/docs/spring-boot-support.md

cherylking:
It has been added here: https://github.com/OpenLiberty/ci.gradle/blob/master/docs/spring-boot-support.md#spring-boot-support

PR #522, opened 2020-08-24
add spring-boot-support docs

dshimo:
for issue #479

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,4 +119,4 @@ Extensions are tasks that improve the compatibility or user experience of third 
 | Extension | Description |
 | --------- | ------------ |
 | [configureArquillian](docs/configureArquillian.md) | Integrates `arquillian.xml` configuration for the Liberty Managed and Remote Arquillian containers in the `liberty-gradle-plugin`. Automatically configures required `arquillian.xml` parameters for the Liberty Managed container. |
-| [Spring Boot Support](docs/spring-boot-support.md#spring-boot-support) | The Liberty Maven Plugin provides support for Spring Boot applications, allowing you to install Spring Boot executable JARs directly to the Liberty runtime. |
+| [Spring Boot Support](docs/spring-boot-support.md#spring-boot-support) | The Liberty Gradle Plugin supports thinning and installing Spring Boot applications onto the Liberty server. |
--- a/README.md
+++ b/README.md
@@ -119,3 +119,4 @@ Extensions are tasks that improve the compatibility or user experience of third 
 | Extension | Description |
 | --------- | ------------ |
 | [configureArquillian](docs/configureArquillian.md) | Integrates `arquillian.xml` configuration for the Liberty Managed and Remote Arquillian containers in the `liberty-gradle-plugin`. Automatically configures required `arquillian.xml` parameters for the Liberty Managed container. |
+| [Spring Boot Support](docs/spring-boot-support.md#spring-boot-support) | The Liberty Maven Plugin provides support for Spring Boot applications, allowing you to install Spring Boot executable JARs directly to the Liberty runtime. |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -119,3 +119,4 @@ Extensions are tasks that improve the compatibility or user experience of third
 | Extension | Description |
 | --------- | ------------ |
 | [configureArquillian](docs/configureArquillian.md) | Integrates `arquillian.xml` configuration for the Liberty Managed and Remote Arquillian containers in the `liberty-gradle-plugin`. Automatically configures required `arquillian.xml` parameters for the Liberty Managed container. |
+| [Spring Boot Support](docs/spring-boot-support.md#spring-boot-support) | The Liberty Maven Plugin provides support for Spring Boot applications, allowing you to install Spring Boot executable JARs directly to the Liberty runtime. |
```
mattbsox, line 122: Change to Liberty Gradle Plugin

docs/spring-boot-support.md
```diff
@@ -0,0 +1,12 @@
+## Spring Boot Support
+
+The springBootUtility command stores the dependent library JARs of the application to the target library cache and packages the remaining application artifacts into a thin application JAR. When you specify a read-only parent library cache, the command creates a target library cache that contains only the libraries that are not available in the parent cache. Use this capability to create efficient Docker layers for your Spring Boot application.
```
cherylking, line 3: I know gradle is different than Maven, but this paragraph is so different from what is provided in liberty-maven-plugin: https://github.com/OpenLiberty/ci.maven/blob/master/docs/spring-boot-support.md#spring-boot-support

Don't we need some of that same info? And an example, and reference to the spring-boot features that need to be configured in server.xml?

mattbsox: Agreed. I think this section should be more along the lines of the Maven docs rather than our Ant docs. 

It'd also be important to mention that the Spring Boot 1.5 plugin is not supported in Gradle versions 5.0 and on.

docs/spring-boot-support.md
```diff
@@ -0,0 +1,35 @@
...
+
+| Parameter | Description | Required |
+| --------  | ----------- | -------  |
+| sourceAppPath | The path of the source application file to thin. | Yes |
+| parentLibCachePath | The directory path of the parent read-only library cache. The parent library cache is searched first to locate existing libraries. If the library is not found, the library is stored in the writable library cache that is specified by the `targetLibCachePath` option. If this option is not specified, no parent library cache is searched. | No |
+| targetLibCachePath | The directory path that is used to save the library cache. If this option is not specified, a `lib.index.cache` directory is created in the parent directory of the source application. | No |
+| targetThinAppPath | The path that is used to save the thin application file. If this option is not specified, a new file is created with the `.spring` extension in the parent directory of the source application. | No |
+
+
+The `server.xml` provided by the `serverXml` parameter should enable the one of the following Spring Boot features.
```
cherylking, line 15: The `server.xml` file provided in the `configDirectory` or `serverXmlFile` parameter....

docs/spring-boot-support.md
```diff
@@ -0,0 +1,35 @@
...
+| ------- | ----------- |
+| springBoot-1.5 | Required to support applications with Spring Boot version 1.5.x. |
+| springBoot-2.0 | Required to support applications with Spring Boot version 2.0.x and above. |
+
+The Liberty features that support the Spring Boot starters can be found [here](https://www.ibm.com/support/knowledgecenter/SSAW57_liberty/com.ibm.websphere.wlp.nd.multiplatform.doc/ae/rwlp_springboot.html). They should be enabled in the `server.xml` along with the appropriate Spring Boot feature.
+
+
+### Gradle Compatibility
+
+There is a known build conflict that Spring Boot Gradle plugin 1.5.x is incompatible with Gradle 5.x. As the Spring Boot 1.5.x plugin won't be updated to support Gradle 5.x, consider upgrading the Spring Boot plugin or downgrading Gradle.
```
cherylking, line 27: Change `won't` to `will not`. I thought 1.5.x worked with Gradle 4.x. We are using the Gradle 4.10 wrapper in our builds for those spring boot tests.

PR #533, opened 2020-09-15
Options dockerRunOpts and dockerfile no longer set option container true.

turkeylurkey:
fixes maven 928


docs/libertyDev.md
```diff
@@ -105,8 +105,8 @@ For the current technology preview, the following limitations apply.
...
-    - Log files: the container will write the Open Liberty log files into the logs directory in the server directory in your project. When running with root permissions dev mode will change the owner of the logs directory and all the files inside to UID 1001 to match the id used by the server in the container. The GID will be 0 (root). When running as a non-root user the set GID bit will be set so that log files will be owned by user 1001 and the group id will be that of the current user.
+    - Configuration files: the Open Liberty server runs on the UID (user identifier) of the current user inside the container. Dev mode makes the server configuration files from your project available inside the container. Therefore dev mode makes all the files in the server directory (`defaultServer`) in your project readable by all users. It also makes all the directories but `workarea` usable by all users.
```
ericglau, line 108: The logical implication is not obvious here.  e.g. Why does the sentence `Dev mode makes the server configuration files from your project available inside the container.` imply `Therefore dev mode makes all the files in the server directory (defaultServer) in your project readable by all users.`?  If it's running as the same UID why do the files need to be readable by all users?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -936,24 +937,6 @@ class DevTask extends AbstractServerTask {
...
         }
-
-        // set container param if dockerfile or dockerRunOpts are set
-        if (!container) {
-            if (dockerfile != null) {
-                if (dockerfile.exists()) {
-                    setContainer(true);
-                    return;
-                } else {
-                    throw new Exception("The file " + dockerfile + " used for dev mode option dockerfile does not exist."
```
ericglau, line 947: Do we perform this check anywhere else to see if the Dockerfile exists?

turkeylurkey: This is checked in ci.common devutil.java line 560.

docs/libertyDev.md
```diff
@@ -105,8 +105,8 @@ For the current technology preview, the following limitations apply.
...
+    - Log files: the container will write the Open Liberty log files into the logs directory in the server directory in your project. When running with root permissions dev mode will change the owner of the logs directory and all the files inside to UID 1001. The GID will be 0 (root). When running as a non-root user the set GID bit will be set so that log files will be owned by current user and the group id will be that of the current user.
```
TrevCraw, line 109: `the set GID bit will be set` - should this just be `the GID bit will be set` ?

docs/libertyDev.md
```diff
@@ -140,6 +140,6 @@ These can also be specified as command line parameters in addition to the ones i
...
-| dockerfile | File | 3.1-M1 (tech preview) | Location of a Dockerfile to be used by dev mode to build the container that runs your server and to specify the context used to build the container image. The default location is `Dockerfile` in the project root. Setting this parameter overrides the `container` parameter to `true`. | No |
+| dockerRunOpts | String | 3.1-M1 (tech preview) | Specifies options to add to the `docker run` command when using dev mode to launch your server in a container. For example, `-e key=value` is recognized by `docker run` to define an environment variable with the name `key` and value `value`. | No |
+| dockerfile | File | 3.1-M1 (tech preview) | Location of a Dockerfile to be used by dev mode to build the container that runs your server and to specify the context used to build the container image. The default location is `Dockerfile` in the project root. | No |
```
TrevCraw, line 144: I know this wasn't a change, but I think we should update the `dockerfile` parameter description:
```
Path to the Dockerfile to be used by dev mode to build the Docker image for the container that will run your Liberty server. The directory containing the Dockerfile will also be the context for the `docker build`. The default path is `Dockerfile` in the project root.
```

@ericglau @turkeylurkey 
thoughts?

ericglau: Sure, but it's not really a path.  If specifying it from the build.gradle, it is a [file](https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:file(java.lang.Object)).

docs/libertyDev.md
turkeylurkey: This code will be deleted but to answer your question the bit actually sets the group id field so it is called the "set gid bit." The wording could have been better but we're deleting this doc too.

docs/libertyDev.md
turkeylurkey: After removing the directory permission code for Linux we can rewrite this part.

docs/libertyDev.md
turkeylurkey: I like the idea to make the first sentence more exact and to split it into two.

ericglau (approved):

PR #552, opened 2020-10-22
Pass in Liberty directory property files

### LINKED ISSUE #551: DevTest and PollingDevTest failures
These tests are failing due to some new code in ci.common that is resolving Liberty directory properties based on the location of the server directory. This does not work for the dev mode scenario as the server.xml has been copied to a temporary directory. The code in question is method initializeLibertyDirectoryPropertyFiles() in ServerFeatureUtil.

Need to determine the locations of the Liberty directory properties, put them in a Map and pass them into ServerFeatureUtil instead. Need to add similar change to ci.maven.

cherylking:
Fixes #551

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,9 @@ before_install:
     - cd ..
     # - echo 'Skip Installing ci.common lib ....'
     - echo 'Installing ci.common lib ....'
-    - git clone https://github.com/OpenLiberty/ci.common.git ./ci.common
+    - git clone https://github.com/cherylking/ci.common.git ./ci.common
     - cd ./ci.common
+    - git checkout passDirectoryProperties
     - mvn clean install
     - cd ..
 script:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -16,8 +16,9 @@ before_install:
     - cd ..
     # - echo 'Skip Installing ci.common lib ....'
     - echo 'Installing ci.common lib ....'
-    - git clone https://github.com/OpenLiberty/ci.common.git ./ci.common
+    - git clone https://github.com/cherylking/ci.common.git ./ci.common
     - cd ./ci.common
+    - git checkout passDirectoryProperties
```
ericglau, line 21: When will this be reverted?  Same question for appveyor.yml

jjvilleg (approved):

.travis.yml
cherylking: I'll revert the yml changes before merging. Thanks for the reminder.

ericglau (approved):

PR #555, opened 2020-10-28
Document COPYing directories and dockerfile limitations requiring manual restart.

turkeylurkey:
Fixes https://github.com/OpenLiberty/ci.maven/issues/982


docs/libertyDev.md
```diff
@@ -126,8 +126,13 @@ For the current technology preview, the following limitations apply.
...
+  - Hot deployment is only supported for individual configuration files that are specified as the source in the Dockerfile's COPY commands. Hot deployment is not supported for COPY commands with entire directories specified as the source. In this case the image it will be rebuilt and the container restarted (regular deployment).
   - Hot deployment is not supported for files or directories copied to the image with the ADD command. When you change any of the files or directories that you ADDed to the image it will be rebuilt and the container restarted.
+  - Certain dockerfile features are not supported by dev mode. In these cases the files are not tracked so if you change these files you must restart the container manually. Type 'r' and press Enter to restart the container.
+   - variable substitution used in the COPY or ADD command e.g. `$PROJECT/config`
+   - wildcards used in the COPY or ADD command e.g. `src/main/liberty/config/*`
+   - paths relative to WORKDIR e.g. `ADD test.txt relativeDir/`
+   - multi-stage builds e.g. `COPY --from=<name>`
```
ericglau, line 135: It seems there are now 4 distinct points that we are trying to convey:
1. For individual files specified, we support hot deployment.
2. For directories, we don't support hot deployment but support automatic rebuild and restart.
3. For ADD commands, we don't support hot deployment but support automatic rebuild and restart.
4. For other Dockerfile features, we don't support hot deployment, we don't support automatic rebuild and restart, but we support manual rebuild and restart.

We should consider how to explain the differences between these points and make them clearer.  Otherwise it is confusing about what is supported or not what the user needs to do in different situations.  If you wish, you can also add a statement to cover https://github.com/OpenLiberty/ci.maven/issues/998.

docs/libertyDev.md
```diff
@@ -126,8 +126,13 @@ For the current technology preview, the following limitations apply.
...
-  - Hot deployment is only supported for individual configuration files that are specified as the source in the Dockerfile's COPY commands. Hot deployment is not supported for COPY commands with variable substitution, wildcard characters, spaces in paths, paths relative to WORKDIR, multi-stage builds, or entire directories specified as the source.
+  - Hot deployment is only supported for individual configuration files that are specified as the source in the Dockerfile's COPY commands. Hot deployment is not supported for COPY commands with entire directories specified as the source. In this case the image it will be rebuilt and the container restarted (regular deployment).
   - Hot deployment is not supported for files or directories copied to the image with the ADD command. When you change any of the files or directories that you ADDed to the image it will be rebuilt and the container restarted.
+  - Certain dockerfile features are not supported by dev mode. In these cases the files are not tracked so if you change these files you must restart the container manually. Type 'r' and press Enter to restart the container.
```
ericglau, line 131: Capitalize `Dockerfile`

docs/libertyDev.md
```diff
@@ -126,8 +126,13 @@ For the current technology preview, the following limitations apply.
...
+   - variable substitution used in the COPY or ADD command e.g. `$PROJECT/config`
+   - wildcards used in the COPY or ADD command e.g. `src/main/liberty/config/*`
+   - paths relative to WORKDIR e.g. `ADD test.txt relativeDir/`
```
ericglau, line 134: I think the point here is using `WORKDIR /my/path` and then using COPY or ADD with a file relative to that WORKDIR instead of to the Docker build context location.

turkeylurkey: Perhaps add a comment like: ADD test.txt relativeDir/ # with WORKDIR /other/project

docs/libertyDev.md
turkeylurkey: ok

docs/libertyDev.md
turkeylurkey: Also I think the section should have a different name.

turkeylurkey: New edit available.

docs/libertyDev.md
turkeylurkey: ok

docs/libertyDev.md
```diff
@@ -101,12 +101,24 @@ The following is a technology preview. The features and parameters described bel
...
 
+The Dockerfile must copy the application .war file and the server configuration files that the application requires into the container. A sample Dockerfile is shown in [Building an application image](https://github.com/openliberty/ci.docker/#building-an-application-image). Note that the context of the `docker build` command used to generate the container image is the directory containing the Dockerfile. When dev mode exits the container is stopped and deleted and the logs are preserved as described.
+
+Dev mode offers different levels of file tracking and deployment depending on the way the file is specified in the Dockerfile. 
+1. When you use the COPY command on an individual file dev mode can track file changes and hot deploy them to the container subject to the limitations below. **This is the most productive way to deploy files.**
```
ericglau, line 111: Add a comma: `When you use the COPY command on an individual file, dev mode can`

docs/libertyDev.md
```diff
@@ -101,12 +101,24 @@ The following is a technology preview. The features and parameters described bel
...
+- E.g. `COPY src/main/liberty/config/server.xml /config/` 
+- Note that the Dockerfile must copy only one .war file for the application.  Other application archive formats or multiple .war files are not supported.
```
ericglau, line 113: Can these be indented so it's more obvious that they are subpoints of (1)?

docs/libertyDev.md
ericglau, line 111: Instead of "most productive" way to deploy files, should we say "recommended"?

docs/libertyDev.md
ericglau, line 113: I don't think we need to say "Other application archive formats or multiple .war files are not supported.".  They are allowed but dev mode will not delete those lines from the temporary Dockerfile.  So if the files are not present at docker build time, the docker build will fail.  But there's nothing particularly wrong with having other formats.

docs/libertyDev.md
```diff
@@ -101,12 +101,24 @@ The following is a technology preview. The features and parameters described bel
...
+2. You can use the COPY command to deploy an entire directory and its sub-directories. In this case, dev mode will detect file changes and use a regular deployment strategy. Dev mode will rebuild the image and restart the container.
```
ericglau, line 114: "regular deployment strategy" is not well defined although I am not too opposed to it.  Another option is to omit the term and combine the last two sentences like, `In this case, dev mode will detect file changes and automatically rebuild the image and restart the container upon changes.`

docs/libertyDev.md
```diff
@@ -101,12 +101,24 @@ The following is a technology preview. The features and parameters described bel
...
+
+Dev mode offers different levels of file tracking and deployment depending on the way the file is specified in the Dockerfile. 
+1. When you use the COPY command on an individual file dev mode can track file changes and hot deploy them to the container subject to the limitations below. **This is the most productive way to deploy files.**
+- E.g. `COPY src/main/liberty/config/server.xml /config/` 
+- Note that the Dockerfile must copy only one .war file for the application.  Other application archive formats or multiple .war files are not supported.
+2. You can use the COPY command to deploy an entire directory and its sub-directories. In this case, dev mode will detect file changes and use a regular deployment strategy. Dev mode will rebuild the image and restart the container. 
+3. The ADD command can be used on individual files including tar files as well as on directories. Again, dev mode will rebuild the image and restart the container.
```
ericglau, line 115: I suggest mentioning that this only happens on file changes, e.g. "Again, dev mode will rebuild the image and restart the container upon file changes. " or "when it detects file changes"

docs/libertyDev.md
```diff
@@ -101,12 +101,24 @@ The following is a technology preview. The features and parameters described bel
...
+
+Dev mode offers different levels of file tracking and deployment depending on the way the file is specified in the Dockerfile. 
+1. When you use the COPY command on an individual file dev mode can track file changes and hot deploy them to the container subject to the limitations below. **This is the most productive way to deploy files.**
+- E.g. `COPY src/main/liberty/config/server.xml /config/` 
+- Note that the Dockerfile must copy only one .war file for the application.  Other application archive formats or multiple .war files are not supported.
+2. You can use the COPY command to deploy an entire directory and its sub-directories. In this case, dev mode will detect file changes and use a regular deployment strategy. Dev mode will rebuild the image and restart the container. 
+3. The ADD command can be used on individual files including tar files as well as on directories. Again, dev mode will rebuild the image and restart the container. 
+4. Certain Dockerfile features are not supported by dev mode. In these cases the files are not tracked so if you change these files you must restart the container manually. **Type 'r' and press Enter to restart the container.**
```
ericglau, line 116: Should we be specific and say it needs to rebuild the image and restart the container?

docs/libertyDev.md
turkeylurkey: kk

docs/libertyDev.md
turkeylurkey: To be honest I assumed the renderer would've done that from the context. I added the spaces.

docs/libertyDev.md
turkeylurkey: Sure, there could be multiple reasons it is recommended.

docs/libertyDev.md
turkeylurkey: Ok, so just multiple war files will lead to problems (they are removed).

docs/libertyDev.md
turkeylurkey: Yes, yours is more to the point.

docs/libertyDev.md
turkeylurkey: /nod

docs/libertyDev.md
ericglau: The reason is the first sentence, which we can link to if possible.  Or call it out like `This is the recommended way to deploy files, so that you can make changes to those files at any time without needing to rebuild the image or restart the container.`

docs/libertyDev.md
turkeylurkey: Yes, that is clearer.

docs/libertyDev.md
```diff
@@ -101,12 +101,24 @@ The following is a technology preview. The features and parameters described bel
...
+
+Dev mode offers different levels of file tracking and deployment depending on the way the file is specified in the Dockerfile. 
+1. When you use the COPY command on an individual file, dev mode can track file changes and hot deploy them to the container subject to the limitations below. **This is the recommended way to deploy files.**
+   - E.g. `COPY src/main/liberty/config/server.xml /config/` 
+   - Note that the Dockerfile must copy only one .war file for the application.  Multiple .war files are not supported.
+2. You can use the COPY command to deploy an entire directory and its sub-directories. In this case, dev mode will detect file changes and automatically rebuild the image and restart the container upon changes.
+3. The ADD command can be used on individual files including tar files as well as on directories. Again, dev mode will rebuild the image and restart the container when it detects file changes. 
+4. Certain Dockerfile features are not supported by dev mode. In these cases the files are not tracked so if you change these files you must rebuild the image and restart the container manually. **Type 'r' and press Enter to restart the container.**
```
ericglau, line 116: How about
> **Type `r` and press Enter to rebuild the image and restart the container.**

turkeylurkey: Ok, I like this. Your previous suggestion was long.

ericglau (approved):

docs/libertyDev.md
turkeylurkey: Ok

docs/libertyDev.md
```diff
@@ -101,12 +101,24 @@ The following is a technology preview. The features and parameters described bel
...
-
 You need to install the Docker runtime (e.g. Docker Desktop) locally to use this Gradle task. You can examine the commands used to build and run the container by viewing the console output of dev mode.
 
 When dev mode runs with container support it still provides the same features. It monitors files for changes and runs tests either automatically or on demand. This mode also allows you to attach a debugger to work on your application. You can review the logs generated by your server in the Liberty directory in your project e.g. build/wlp/usr/servers/defaultServer/logs.
 
+The Dockerfile must copy the application .war file and the server configuration files that the application requires into the container. A sample Dockerfile is shown in [Building an application image](https://github.com/openliberty/ci.docker/#building-an-application-image). Note that the context of the `docker build` command used to generate the container image is the directory containing the Dockerfile. When dev mode exits, the container is stopped and deleted, and the logs are preserved as described.
```
TrevCraw, line 108: Where do we describe the logs being preserved?

PR #573, opened 2020-12-07
Github actions for build

cherylking:
Fixes #573 

Adding the following builds via github actions:

- windows-latest and Ubuntu-18.04
- Java 8 and 11
- Versions 20.0.0.9 and 20.0.0.12 for ol (OpenLiberty) and wlp (WebSphere Liberty)

Omitting Java 8 on 20.0.0.9 to reduce length of builds.

Also, only testing the SpringBoot 15 tests on the Gradle 4.10 wrapper with Ubuntu. I encountered various problems running them on Windows (out of disk space, no such method found, etc).

Finally, I had to change the groupId for one of our dependencies. I do not know how it was working before, as it was using the wrong group id (api vs spi).


mattbsox (approved):
Looks good.

PR #585, opened 2021-02-05
Skip classpath since not needed for gradle recompiler

ericglau:
Part of https://github.com/OpenLiberty/ci.maven/issues/486

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,7 @@ compileTestGroovy {
 }
 
 def libertyAntVersion = "1.9.7"
-def libertyCommonVersion = "1.8.13"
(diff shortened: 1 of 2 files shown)


turkeylurkey (commented):
How come Maven has "compileArtifactPaths.addAll(project.getCompileClasspathElements());" but Gradle doesn't have something similar?
Edit: I see that Gradle sets useBuildRecompile and Maven does not. I presume that is the answer?

turkeylurkey (approved):

ericglau:
@turkeylurkey Yes, Gradle has useBuildRecompile so dev mode does not need to deal with the classpaths.  This is why we can pass null as the artifact paths to DevUtil.watchFiles().

PR #586, opened 2021-02-19
adding support for user feature install

hermankailey:
adding support for user feature install via specified featuresBom and to parameter


src/main/groovy/io/openliberty/tools/gradle/tasks/InstallFeatureTask.groovy
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corporation 2014, 2020.
+ * (C) Copyright IBM Corporation 2014, 2020, 2021.
```
mattbsox, line 2: Change this to `2014, 2021.`

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractFeatureTask.groovy
```diff
@@ -37,16 +39,22 @@ public class AbstractFeatureTask extends AbstractServerTask {
...
     private class InstallFeatureTaskUtil extends InstallFeatureUtil {
-        public InstallFeatureTaskUtil(File installDir, String from, String to, Set<String> pluginListedEsas, List<ProductProperties> propertiesList, String openLibertyVerion, String containerName)  throws PluginScenarioException, PluginExecutionException {
-            super(installDir, from, to, pluginListedEsas, propertiesList, openLibertyVerion, containerName)
+		
+		
+        public InstallFeatureTaskUtil(File installDir, String from, String to, Set<String> pluginListedEsas, List<ProductProperties> propertiesList, String openLibertyVerion, String containerName, List<String> additionalJsons)  throws PluginScenarioException, PluginExecutionException {
+            super(installDir, from, to, pluginListedEsas, propertiesList, openLibertyVerion, containerName, additionalJsons)
         }
+		
+		public Project
```
mattbsox, line 57: Think you can remove this line.

src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corporation 2014, 2020.
+ * (C) Copyright IBM Corporation 2014, 2020, 2021.
```
mattbsox, line 2: Change this to `2014, 2021.`

src/main/groovy/io/openliberty/tools/gradle/LibertyTaskFactory.groovy
```diff
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corporation 2018, 2020.
+ * (C) Copyright IBM Corporation 2018, 2020, 2021.
```
mattbsox, line 2: Change this to `2018, 2021.`

src/main/groovy/io/openliberty/tools/gradle/tasks/PrepareFeatureTask.groovy
```diff
@@ -0,0 +1,48 @@
+/**
+ * (C) Copyright IBM Corporation 2014, 2020, 2021.
```
cherylking, line 2: I don't think 2014 should be listed in this copyright.

cherylking (changes requested):
Note that the gradle.yml needs updating before this can be merged. Also, I do not want this to go into the next release we have planned for next week.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractFeatureTask.groovy
```diff
@@ -185,8 +204,26 @@ public class AbstractFeatureTask extends AbstractServerTask {
...
+
+				// need this for gradle to evaluate the project
+				// and load the different plugins and extensions
+				newProject.evaluate();
+			} catch (Exception e) {
+				logger.error("Could not parse build.gradle " + e.getMessage());
+				logger.debug('Error parsing build.gradle', e);
+			}
+		}
+		createNewInstallFeatureUtil(pluginListedEsas, propertiesList, openLibertyVerion, containerName, additionalJsons)
```
cherylking, line 226: If the above section fails (gets an exception), we are allowing it to continue on? Will it likely fail?

jjiwooLim: Added return statement when the exception is caught to make it fail.

cherylking: If you are going to return, shouldn't it `return null` ?

cherylking: Actually, the caller does not check the return value and tries to call a method on it. I think the exception needs to be propagated rather than eaten here.

jjiwooLim: Yeah I was just going over that, I will make it throw PluginExecutionException instead.

cherylking:
This is superseded by #622

PR #588, opened 2021-02-26
InstallLiberty will re-run if config properties have been changed.

### LINKED ISSUE #221: installLiberty task reports UP-TO-DATE when liberty version changes
Recently I updated a project to use the latest beta:

```js
    liberty {
        install {
            // use 1 liberty install for the whole repo
            baseDir = rootProject.buildDir
            runtimeUrl = "https://public.dhe.ibm.com/ibmdl/export/pub/software/websphere/wasdev/downloads/wlp/beta/wlp-beta-2018.2.0.0.zip"
        }
    }
```

And got a failure when features present in the new beta could not be found by the runtime or installed by installUtility.  Turns out this is because the `installLiberty` task is considered to be up-to-date if `liberty.install.baseDir` exists, but it does not check to make sure it's the right version of liberty installed.

The `libertyInstall` task should be updated to compare product versions of the maven/runtimeUrl dependency to the product version in `${liberty.install.baseDir}/lib/versions/WebSphereApplicationServer.properties` to ensure the runtime itself is up to date.

jjvilleg:
Fixes #221

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/InstallLibertyTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/InstallLibertyTask.groovy
@@ -96,6 +96,12 @@ class InstallLibertyTask extends AbstractTask {
         return project.liberty.install.type
     }
 
+    @Input
+    @Optional
+    Properties getLibertyGeneralRuntimeProperties_() {
+        return project.liberty.runtime
+    }
+
 
     @TaskAction
     void install() {
@@ -249,6 +255,7 @@ class InstallLibertyTask extends AbstractTask {
         }
 
         result.put('offline', project.gradle.startParameter.offline)
+        result.put('skipAlreadyInstalledCheck', "true")
 
         return result
     }
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/InstallLibertyTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/InstallLibertyTask.groovy
@@ -59,25 +59,57 @@ class InstallLibertyTask extends AbstractTask {
         return project.configurations.libertyRuntime
     }
 
+    @Input
+    @Optional
+    String getLibertyRuntimeUrl() {
+        return project.liberty.install.runtimeUrl
+    }
+
+    @Input
+    @Optional
+    String getLibertyLicenseCode() {
+        return project.liberty.install.licenseCode
+    }
+
+    @Input
+    @Optional
+    String getLibertyVersion() {
+        return project.liberty.install.version
+    }
+
+
+    @Input
+    @Optional
+    String getLibertyUsername() {
+        return project.liberty.install.username
+    }
+
+    @Input
+    @Optional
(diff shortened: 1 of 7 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/InstallLibertyTask.groovy
```diff
@@ -59,25 +59,57 @@ class InstallLibertyTask extends AbstractTask {
         return project.configurations.libertyRuntime
     }
 
+    @Input
+    @Optional
+    String getLibertyRuntimeUrl() {
+        return project.liberty.install.runtimeUrl
+    }
```
cherylking, line 66: What about the project.liberty.runtime properties?
Screen Shot 2021-02-26 at 3 17 17 PM

cherylking: That is in the liberty extension. We also support gradle properties for those.
Screen Shot 2021-02-26 at 3 21 01 PM

src/main/groovy/io/openliberty/tools/gradle/tasks/InstallLibertyTask.groovy
```diff
@@ -59,25 +59,57 @@ class InstallLibertyTask extends AbstractTask {
...
-            String licenseFilePath = project.configurations.getByName('libertyLicense').getAsPath()
-            if (licenseFilePath) {
-                def command = "java -jar " + licenseFilePath + " --acceptLicense " + project.buildDir
-                def process = command.execute()
-                process.waitFor()
-            }
-            
-        } else {
-            logger.info ("Liberty is already installed at: " + getInstallDir(project))
+        def params = buildInstallLibertyMap(project)
```
cherylking, line 102: So why don't we need some check here? This won't get called if it was already installed?

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/InstallLibertyTask.groovy
jjvilleg: Exactly. Even if the Inputs weren't up to date, this check would keep it from reinstalling WLP with the updated configuration. We are already adding a checking if the contents are there in this line
https://github.com/OpenLiberty/ci.gradle/blob/master/src/main/groovy/io/openliberty/tools/gradle/tasks/InstallLibertyTask.groovy#L51

cherylking (commented):
So the changes look good, but we probably need a couple of tests to prove it works correctly.

cherylking:
Looks like your changes are causing this test to fail (invalid installDir location):
https://github.com/OpenLiberty/ci.gradle/pull/588/checks?check_run_id=2000509793#step:9:10984

cherylking:
And another test failure: https://github.com/OpenLiberty/ci.gradle/pull/588/checks?check_run_id=2000509793#step:9:19127

cherylking (approved):

PR #595, opened 2021-04-23
Fix Gradle 7.0 issues

### LINKED ISSUE #592: Identify and fix issues related to Gradle 7 support
A customer reported the following error when trying to run libertyDev task with Gradle 7:
Type 'DevTask' property 'container' is missing an input or output annotation.

The fix for that particular issue can be found in this PR: #590 

However, we would need to identify and fix all issues related to Gradle 7 support, not just this one. @ericglau found the following when attempting to run with Gradle 7:

```
A problem was found with the configuration of task ':installLiberty' (type 'InstallLibertyTask').
  - Type 'InstallLibertyTask' property 'libertyRuntimeConfiguration' has @Input annotation used on property of type 'Configuration'.
    Reason: A property of type 'Configuration' annotated with @Input cannot determine how to interpret the file.
    Possible solutions:
      1. Annotate with @InputFile for regular files.
      2. Annotate with @InputDirectory for directories.
      3. If you want to track the path, return File.absolutePath as a String and keep @Input.
    Please refer to https://docs.gradle.org/7.0-rc-2/userguide/validation_problems.html#incorrect_use_of_input_annotation for more details about this problem.
```
We currently support Gradle 6 and should look into these Gradle 7 issues. Currently milestone 1 is available for Gradle 7.

mattbsox:
- Fixed issues when running the plugin with Gradle 7.0
- Updated the Gradle wrapper to the latest Gradle 6.X version: 6.8.3
- Updated Spring Boot plugin version to 2.4.3 from 2.0.4 in the Spring Boot 2.0 tests
- DevTest and PollingDevTest failing when run on Gradle 7.0. Underlying issue seems to be an AbstractMethodError when trying to access ServerExtension properties during project/task configuration. The `libertyDev` task seems to work properly when run outside of the test files.

The current version of the Spring Boot plugin used in the Spring Boot 2.0 tests requires Gradle 5 (5.6.x only) or Gradle 6 (6.3 or later)

Fixes #592


src/test/resources/arquillian-tests/skipWithXmlManaged/build.gradle
```diff
@@ -16,7 +16,7 @@ buildscript {
 project.buildDir = '../build'
 
 dependencies {
-    implementation 'io.openliberty.arquillian:arquillian-liberty-managed:1.0.2'
+    implementation 'io.openliberty.arquillian:arquillian-liberty-managed:1.0.6'
```
cherylking, line 19: I just released 1.0.7 today. So might as well update these now.

src/main/groovy/io/openliberty/tools/gradle/LibertyTasks.groovy
```diff
@@ -38,7 +38,9 @@ public class LibertyTasks {
         project.libertyRun {
             dependsOn 'libertyCreate'
 
-            if (dependsOnApps(project.liberty.server)) dependsOn 'deploy'
+            if (dependsOnApps(project.liberty.server)) {
```
cherylking, line 41: General comment to update the copyright year in all the groovy files you updated.

cherylking (changes requested):

cherylking (approved):

cherylking:
I like the test case fix. Did you see the new issue opened about the Windows builds failing with the idea of running gradle clean task at end of integration tests? Would that be worth a try?

mattbsox:
Yeah, that makes sense. I'll see if there is a good way to tie that to CI builds with a flag somewhere. I'd only like to do that when necessary.

PR #602, opened 2021-05-07
Added output to libertyPackage

### LINKED ISSUE #472: Cannot programmatically get output file from libertyPackage task
I'm unable to get access to the output of the libertyPackage task using the task's `outputs` property as is standard with most gradle tasks. I suspect this is because the task is missing logic to declare its outputs. Adding this might also ensure that gradle's up-to-date logic works properly as well. This issue was discovered with the following line of gradle code:
```
libertyPackage.outputs.files.singleFile
```
Which produces this error:
```
Expected task 'libertyPackage' output files to contain exactly one file, however, it contains no files.
```

mattbsox:
- The file generated by `libertyPackage` is now an output for the task
- Marked `isUserDirSpecified()` as an internal method, fixes a Gradle 7 issue

Fixes #472

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/PackageTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/PackageTask.groovy
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corporation 2014, 2019.
+ * (C) Copyright IBM Corporation 2014, 2021.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/PackageTask.groovy
```diff
@@ -15,6 +15,7 @@
  */
 package io.openliberty.tools.gradle.tasks
 
+import org.gradle.api.tasks.OutputFile
```
cherylking, line 18: Update copyright year.

cherylking (approved):

Issue #607, opened 2021-06-14
command line property -Pliberty.runtime.version does not override default OL runtime version


mezarin:
When I run this command that defines the -Pliberty.runtime.version property (For example: gradle libertyDev -Pliberty.runtime.version=21.0.0.3), the Liberty version installed is the latest one and not the version specified. 
This happens whether or not I specify the libertyRuntime dependency in build.gradle.

According to this documentation: https://github.com/OpenLiberty/ci.gradle/blob/main/docs/installLiberty.md, and the example command above, OL version 21.0.0.3 should have been resolved/installed.

I can workaround this by having a gradle.properties file with the above property specified, or I can define/override the liberty version directly in build.gradle (i.e. under dependencies or liberty.runtime).

cherylking:
@ericglau The `InstallLibertyTask` looks for the `liberty.runtime` project properties in `buildInstallLibertyMap()`. Does the `DevTask` need to do something to pass those along to the `BuildLauncher` that calls `libertyCreate`?

PR #608, opened 2021-06-15
Pass liberty.runtime properties to libertyCreate

ericglau:
Fixes https://github.com/OpenLiberty/ci.gradle/issues/607


ericglau:
If Liberty runtime properties are defined from command line like `-Pliberty.runtime.version=21.0.0.3` or in gradle.properties like `liberty.runtime.version=21.0.0.3`, then this PR will include them in the arguments for libertyCreate when called from libertyDev

If Liberty runtime properties are defined in build.gradle as an array like:
```
liberty {
    runtime = ['version':'21.0.0.3']
}
```
then those will be picked up automatically without dev mode needing to pass it in to libertyCreate.

cherylking (approved):

PR #637, opened 2021-11-01
bump ant, common, and release versions

dshimo:


build.gradle
```diff
@@ -10,7 +10,7 @@ apply plugin: "com.gradle.plugin-publish"
 
 archivesBaseName = 'liberty-gradle-plugin'
 group = 'io.openliberty.tools'
-version = '3.2.2-SNAPSHOT'
+version = '3.2.3-SNAPSHOT'
```
cherylking, line 13: This was supposed to be 3.3-SNAPSHOT. New functionality warrants a point release instead of a minor fix release.

cherylking (changes requested):

build.gradle
cherylking: Also, you will need to rebase off the latest master to get the tests to pass.

Issue #629, opened 2021-10-08
Allow installDir property to be set on the CLI


awisniew90:
The installDir property allows you to specify an existing Liberty installation. Currently this is only configurable via your build.gradle file, but would be beneficial to have it set on the command line. Specifically, we'd like to leverage this capability in the devfile application stack which has Liberty pre-installed inside a container.

PR #651, opened 2021-11-18
Add installDir project property detection

dshimo:
for issue #629


mattbsox (approved):

src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
```diff
@@ -169,7 +169,11 @@ class Liberty implements Plugin<Project> {
     }
 
     private static File getInstallDir(Project project) {
-        if (project.liberty.installDir == null) {
+        if (project.hasProperty('liberty.installDir')) {
```
cherylking, line 172: Two things...I think we should call `checkAndAppendWlp(project)` here...and that method needs updating to look for the property before using `project.liberty.installDir`.  Or make a second `checkAndAppendWlp(String)` and pass in the path to check.

src/test/groovy/io/openliberty/tools/gradle/InstallLiberty_installDir_missing_wlp_Test.groovy
```diff
@@ -45,4 +45,16 @@ class InstallLiberty_installDir_missing_wlp_Test extends AbstractIntegrationTest
         String output = result.getOutput()
         assert output.contains("path does not reference a wlp folder") : "Expected warning about installDir path not containing wlp"
     }
+
+    @Test
+    void test_installLiberty_installDir__cli_property() {
+        BuildResult result = GradleRunner.create()
+            .withProjectDir(buildDir)
+            .forwardOutput()
+            .withArguments('installLiberty', '-Pliberty.installDir=installDir-valid-install/build/wlp', '-i', '-s')
```
cherylking, line 54: After adding the logic to check the path above, you can add another test here to pass in a path with `-Pliberty.installDir` that does not end with `/wlp`.

cherylking (changes requested):

cherylking (approved):

docs/libertyExtensions.md
```diff
@@ -14,7 +14,7 @@ The [deploy](deploy.md), [undeploy](undeploy.md), [libertyPackage](libertyPackag
...
 | baseDir | String | 3.0 | The base installation directory. The actual installation directory of WebSphere Liberty server will be `${baseDir}/wlp`. The default value is `${project.buildDir}`. This was moved from the properties in the `install` block in version 3.0.| No |
 | cacheDir | String | 3.0 | The directory used for caching downloaded files such as the license or `.jar` files. The default value is `${java.io.tmpdir}/wlp-cache`. This was moved from the properties in the `install` block in version 3.0.| No |
-| installDir | String | 1.0 | Path to the WebSphere Liberty server installation `wlp` directory. To use a pre-installed version of Liberty, set this property to the path of the Liberty `wlp` directory, including `wlp` in the path. | No |
+| installDir | String | 1.0 | Path to the WebSphere Liberty server installation `wlp` directory. To use a pre-installed version of Liberty, set this property to the path of the Liberty `wlp` directory, including `wlp` in the path. <br /> <br /> Additionally, `installDir` can be specified in the `gradle.properties` file or from the command line using `-Pliberty.installDir={wlp}`. | No |
```
cherylking, line 17: I don't think I would put `{wlp}` in the example, and let's remove the break.

cherylking: Also, installLiberty.md might be a good place to show an example usage of the installDir property. That is where the `runtime` properties for specifying groupId, artifactId and version for the Liberty runtime are discussed in more detail.

cherylking (approved):

PR #657, opened 2021-11-23
Add a system where the binary scanner handler detects an error generated

turkeylurkey:
by the scanner and calls it again with fewer restrictions in order to
generate a list of suggested features that should work. This error
and suggested solution is then displayed to the user.

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
@@ -299,6 +299,7 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
                     String problemMessage = scannerException.getMessage();
                     if (problemMessage == null || problemMessage.isEmpty()) {
                         logger.debug("RuntimeException from binary scanner without descriptive message", scannerException);
+                        logger.error("Error scanning the application for Liberty features.");
                     } else {
                         Set<String> conflicts = parseScannerMessage(problemMessage);
                         Set<String> sampleFeatureList = null;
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
```diff
@@ -48,7 +49,21 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
...
+            "Java EE, or Jakarta EE. Refer to the following set of suggested features for guidance: %s";
+    private static final String BINARY_SCANNER_CONFLICT_MESSAGE2 = "A working set of features could not be generated due to conflicts " +
+            "between configured features: %s. Review and update your server configuration to ensure it is not using conflicting " +
+            "features from different levels of MicroProfile, Java EE, or Jakarta EE. Refer to the following set of " +
+            "suggested features for guidance: %s";
+    private static final String BINARY_SCANNER_CONFLICT_MESSAGE3 = "A working set of features could not be generated due to conflicts " +
+            "in the application’s API usage: %s. Review and update your application to ensure it is not using conflicting APIs " +
+            "from different levels of MicroProfile, Java EE, or Jakarta EE.";
+    private static final String BINARY_SCANNER_CONFLICT_MESSAGE4 = "[None available]"; // format should match JVM Set.toString()
+
```
kathrynkodama, line 64: Should we consider putting these strings in a ci.common util class since they are the same across ci.maven/ci.gradle? Can be a future enhancement

kathrynkodama (approved):

src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
```diff
@@ -301,4 +397,80 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
...
+        }
+        public Set<String> getConflicts() {
+            return conflicts;
+        }
+    }
+
+    // A class that encapsulates a list of conflicting features, a suggested list of replacements
+    // and a flag that indicates whether the conflicts were found in the features existing in the
+    // app's server config or if the conflicts exist in the binary files we examined.
+    private class RecommendationSetException extends Exception {
```
kathrynkodama, line 456: We could also move these Exception classes to a util class in ci.common eventually

src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
turkeylurkey: Yes, I'm going to work on a binary scanner handler class.

src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
turkeylurkey: Agreed.

PR #660, opened 2021-11-30
Use the binary scanner support in ci.common

turkeylurkey:
Hopefully no change in behaviour except to get the EE and MicroProfile version from build file if umbrella feature is used. Matches the support in ci.maven. If we want to change how the EE/MP version is set or obtained we should probably track that in a separate issue and fix in both LMP and LGP.
Fixes https://github.com/OpenLiberty/ci.maven/issues/1309

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
@@ -33,6 +33,7 @@ import org.xml.sax.SAXException
 import io.openliberty.tools.common.plugins.util.InstallFeatureUtil
 import io.openliberty.tools.common.plugins.config.ServerConfigXmlDocument
 import io.openliberty.tools.common.plugins.config.XmlDocument
+import io.openliberty.tools.common.plugins.util.BinaryScannerUtil
 import io.openliberty.tools.common.plugins.util.DevUtil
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
```diff
@@ -399,86 +257,73 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
...
-            return conflicts;
+        @Override
+        public void warn(String msg) {
+            logger.warn(msg);
         }
-        public Set<String> getSuggestions() {
-            return suggestions;
+        @Override
+        public void info(String msg) {
+            logger.info(msg);
```
ericglau, line 326: We may want to change this to `logger.lifecycle(msg);` similar to DevTask https://github.com/OpenLiberty/ci.gradle/blob/1bca3270867ddfffa0e8654479dd420b0764c363/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy#L327 so that Gradle will show these by default.

ericglau (approved):

src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
turkeylurkey: 👍

PR #661, opened 2021-12-02
Get existing features for feature gen from the source config dir

kathrynkodama:
See https://github.com/OpenLiberty/ci.maven/pull/1312

Have generate features get existing user specified features from the source directory rather than the target directory.

Added `ServerFeatureTaskUtil` to `AbstractFeatureTask`. 
To reduce code duplication and have dev mode make use of `ServerFeatureTaskUtil`:
`DevTask` now extends `AbstractFeatureTask` which extends `AbstractServerTask`
(previously was: `DevTask` extends `AbstractServerTask`)
 
### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractFeatureTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractFeatureTask.groovy
@@ -185,14 +185,9 @@ public class AbstractFeatureTask extends AbstractServerTask {
         return featuresToInstall
     }
 
-    private void createNewServerFeatureUtil() {
-        servUtil = new ServerFeatureTaskUtil();
-        return;
-    }
-
     protected ServerFeatureUtil getServerFeatureUtil() {
         if (servUtil == null) {
-            createNewServerFeatureUtil();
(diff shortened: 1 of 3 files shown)


ericglau (commented):
Should we omit the `.idea` and `.vscode` folders?

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractFeatureTask.groovy
```diff
@@ -145,14 +182,26 @@ public class AbstractFeatureTask extends AbstractServerTask {
         }
 
         Set<String> featuresToInstall = InstallFeatureUtil.combineToSet(pluginListedFeatures, dependencyFeatures, serverFeatures)
-        return featuresToInstall 
+        return featuresToInstall
+    }
+
+    private void createNewServerFeatureUtil() {
+        servUtil = new ServerFeatureTaskUtil();
+        return;
```
ericglau, line 190: The `return` statement isn't strictly needed here.  Then since this is one line, might as well just put it directly in line 195.

ericglau (approved):

ericglau (approved):

turkeylurkey (approved):

PR #685, opened 2022-01-06
Fix timeout cast exception

### LINKED ISSUE #684: Property timeout causes a class cast exception if is set
If the timeout property is available to override the 30s for waiting for the deploy to be done, i tracked down to the task DeployTask  on line https://github.com/OpenLiberty/ci.gradle/blob/0b72dc4f92df09fbb9896f35d230bb503db9621c/src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy#L646, the property is defined as String but it's used as long, should this property used on liberty start as well why exist two properties to do the same thing timeout and verifyAppStartTimeout

mattbsox:
Fixes #684

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
@@ -924,7 +924,7 @@ class DevTask extends AbstractFeatureTask {
         }
 
         if (serverStartTimeout == null) {
-            if (server.timeout != null && server.timeout.isInteger()) {
+            if (server.timeout != null && !server.timeout.isEmpty()) {
                 try {
                     serverStartTimeout = Integer.valueOf(server.timeout);
                 } catch (NumberFormatException e) {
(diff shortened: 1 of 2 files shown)


cherylking (commented):
Should we check that it is a long before calling `longValue()`? We have code in DevTask that checks `isInteger()` before calling `Long.valueOf()` on the `server.timeout`.

mattbsox:
Yeah, I'll add a check for that.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -925,7 +925,12 @@ class DevTask extends AbstractFeatureTask {
 
         if (serverStartTimeout == null) {
             if (server.timeout != null && server.timeout.isInteger()) {
-                serverStartTimeout = server.timeout.toInteger();
+                try {
```
cherylking, line 928: I don't believe `isInteger()` is valid on the line before this either. Probably should just check that the string is notEmpty().

cherylking (changes requested):

cherylking (approved):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
mattbsox: Yeah, missed that. Fixed it.

PR #692, opened 2022-01-14
Add generate features test. Rewrite log verification routine.

turkeylurkey:


src/test/groovy/io/openliberty/tools/gradle/DevTest.groovy
```diff
@@ -145,8 +157,8 @@ class DevTest extends AbstractIntegrationTest {
     @Test
     /* simple double check. if failure, check parse in ci.common */
     public void verifyJsonHost() throws Exception {
-        checkLogMessage(2000, "CWWKT0016I");   // Verify web app code triggered
-        checkLogMessage(2000, "http:\\/\\/");  // Verify escape char seq passes
+        assertTrue(verifyLogMessage(2000, "CWWKT0016I"));   // Verify web app code triggered
+        // TODO assertTrue(verifyLogMessage(2000, "http:\\/\\/"));  // Verify escape char seq passes
```
kathrynkodama, line 161: Why was this commented out?

turkeylurkey: Two things: the previous code was not testing anything so I added asserts to check the output of checkLogMessage. Next the same test in Maven is commented out so I understand the check is not valid and needs investigation.

kathrynkodama: Makes sense, thank you

kathrynkodama (approved):

PR #696, opened 2022-01-18
Catch the FeatureModifiedException from the binary scanner and report

turkeylurkey:
the feature conflicts or pass on the new feature names if ok.


turkeylurkey:
Most tests pass and same code approved in ci.maven.

PR #698, opened 2022-01-24
Add generateFeatures standalone integration test

kathrynkodama:
Basic framework for `generateFeatures` task integration tests.


turkeylurkey (approved):
I have the same comment as Maven regarding the before/after class method possibly being done before each specific test but that could be reworked when there is more than one test.

Issue #665, opened 2021-12-07
Generate features, improve error message when class files do not exist 


kathrynkodama:
Running `gradle generateFeatures` on a project without any class files, a StackOverflow error is returned:
```
kathrynkodama@MacBook-Pro demo-devmode-scan % gradle generateFeatures

> Task :generateFeatures
Parsing the server file /Users/kathrynkodama/devex/featureGeneration/demo-devmode-scan/src/main/liberty/config/server.xml1 for features and includes.
Parsing the server file /Users/kathrynkodama/devex/featureGeneration/server.xml for features and includes.
null

> Task :generateFeatures FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':generateFeatures'.
> java.lang.StackOverflowError (no error message)

```

Improve this error message to indicate that users should run `gradle compileJava` first.

turkeylurkey:
Similarly, if there is no classes directory at all the scanner can return an exception:
`java.lang.IllegalArgumentException: CWMIG12026E: The /Users/paulg/WAS/devExperience/featureGeneration/demo-devmode/build/classes/java/main file that is specified by the binaryInputPath parameter does not exist.` We should check it exists before we call the scanner.

PR #700, opened 2022-01-25
Feature generation, display warning when class files do not exist

kathrynkodama:
Fixes https://github.com/OpenLiberty/ci.gradle/issues/665
See https://github.com/OpenLiberty/ci.common/pull/342

When class directories or files do not exist, log a warning to users. Logging as warning so that conflicts between users specified features will still be detected by the binary scanner and will not block startup of dev mode.

`gradle compileJava` differs slightly in behaviour from `mvn compile`. `gradle compileJava` will not create a `build/classes` directory if the `src/main/java` directory does not exist OR if the `src/main/java` directory does exist but does not contain any classes.


turkeylurkey:
Hey I got zenhub installed

turkeylurkey (approved):

src/test/groovy/io/openliberty/tools/gradle/GenerateFeaturesTest.groovy
```diff
@@ -85,6 +89,20 @@ class GenerateFeaturesTest extends AbstractIntegrationTest {
...
+    public void noClassFiles() throws Exception {
+        // do not compile before running generateFeatures
+        runProcess(" generateFeatures");
+
+        // verify that generated features file was not created
+        assertFalse(newFeatureFile.exists());
+
+        // verify class files not found warning message
+        assertTrue(processOutput.contains("Could not find class files to generate features against. "
+                + "Liberty features will not be generated. Ensure your project has first been compiled."));
```
TrevCraw, line 102: We should make this warning a constant so we are not checking a String that may change. Probably best to have it in ci.common to use across Maven and Gradle as well.

kathrynkodama: The message varies very slightly in wording between Maven and Gradle (class directories vs class files) so I don't think it should go in ci.common, but I created a constant in GenerateFeaturesTask

PR #702, opened 2022-01-26
Enable Windows testing with GitHub Actions + Log Artifacts

dshimo:
Moved contents from D:/ drive to C:/ drive to avoid limited disk space issue while testing on Windows
Exclude LibertyTest from Windows testing due to noticed hanging issue during `run`. Test is run on Ubuntu
Corrected previous installDir property logging to now be consistent with OS

**Log Artifacts**:
Whenever a build fails GitHub actions will now upload a downloadable zip of the Gradle reports page, since reviewing the logs in the console isn't the most readable.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -10,15 +10,37 @@ on:
     branches: [ main ]
 
 jobs:
-  build:
+  cache:
+    name: Cache Maven and Gradle packages
     runs-on: ${{ matrix.os }}
-
-    # Steps represent a sequence of tasks that will be executed as part of the job
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


cherylking (approved):
I like it! I was able to download the build report for the one failing build and see which test/method failed. And 10 out of 11 builds passed. So much better!

kevinwlau (commented):
With the cache step moved out into a separate build, it seems like the builds are no longer using the cache?

If I understand correctly from a previous discussion, it seems the Windows builds aren't utilizing the cache. So maybe the caching can be moved into a step in `build-unix`

cherylking:
@dshimo Can you answer Kevin's questions in the PR? Want to make sure they were considered before merging the PR.

PR #703, opened 2022-01-27
Cache changes

kevinwlau:
- Move caching steps back into the builds
- Switch back to official actions/setup-java now that it has the jdk we use
- Change the copy build reports step to only run if build failed (have not been able to test with a build failure yet)

To make the yaml file slightly easier to maintain, I consolidated the Windows/Unix builds back into one, with OS qualifiers on the OS-specific steps. This way, each step of the build, regardless of OS, is grouped together.

However, it's not as nicely separated on the Actions UI now.

With the recent change to separate Unix/Windows builds: 
image

Back to single consolidated build (which we've used all this time before):
image

With the consolidation, the steps during the build looks messier, as it shows the skipped steps that are meant for the other OS:
image

Any opinions on the consolidated vs separate OS builds?

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -10,24 +10,20 @@ on:
     branches: [ main ]
 
 jobs:
-  build:
-    runs-on: ${{ matrix.os }}
+# UNIX BUILDS
+  build-unix:
+    runs-on: ubuntu-18.04
     strategy:
       fail-fast: false
       matrix:
         # test against latest update of each major Java version, as well as specific updates of LTS versions:
-        os: [ubuntu-18.04, windows-latest]
         RUNTIME: [ol, wlp]
         RUNTIME_VERSION: [21.0.0.3, 20.0.0.12]
         java: [11, 8]
         exclude:
         - java: 8
           RUNTIME_VERSION: 20.0.0.12
-        - java: 8
-          RUNTIME: wlp
-          os: windows-latest
-
-    name: ${{ matrix.RUNTIME }} ${{ matrix.RUNTIME_VERSION }}, Java ${{ matrix.java }}, ${{ matrix.os }}
+    name: ${{ matrix.RUNTIME }} ${{ matrix.RUNTIME_VERSION }}, Java ${{ matrix.java }}, Linux
     steps:
 # Checkout repos
     - name: Checkout ci.gradle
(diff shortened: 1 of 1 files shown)


cherylking:
I prefer the separation over the consolidation.

.github/workflows/gradle.yml
```diff
@@ -49,96 +25,108 @@ jobs:
...
         repository: OpenLiberty/ci.ant
         path: ci.ant
-    - name: Install ci.ant and ci.common
+    # Moving and cloning to C: drive for Windows for more disk space
+    - name: Clone ci.ant, ci.common, ci.gradle repos to C drive (Windows)
+      if: ${{ runner.os == 'Windows' }}
+      run: |
+        cp -r D:/a/ci.gradle/ci.gradle C:/ci.gradle
+        git clone https://github.com/OpenLiberty/ci.common/ C:/ci.common
+        git clone https://github.com/OpenLiberty/ci.ant C:/ci.ant
```
cherylking, line 59: Why does one https path end with `/` and the other one doesn't? Also, when I do a git clone with https it usually ends with `.git`. Why is this different?

cherylking (commented):
Just one question/comment

.github/workflows/gradle.yml
kevinwlau: I didn't even notice either of these while rearranging items.

I think the trailing slash doesn't matter at the end of these URLs, but I can clean it up.
and git will automatically resolve the repository, and will look for the `.git` version if it can't find it.

dshimo: I added these. It was simple oversight on both comments. I manually typed the git URLs from memory and forgot to correct them. 
Although it performs the same, it would be more correct to clone with the `.git` extension

dshimo (approved):

.github/workflows/gradle.yml
```diff
@@ -91,41 +70,97 @@ jobs:
...
+# Install dependencies
+    - name: Install ci.ant and ci.common
+      run: |
+        cd C:/ci.ant
+        mvn clean install
+        cd C:/ci.common
+        mvn clean install
+# Run tests
+    - name: Run tests with Gradle on Windows
+      if: ${{ runner.os == 'Windows' }}
```
cherylking, line 150: Is this check needed? Think we are in a Windows only section here.

cherylking (changes requested):
One comment

.github/workflows/gradle.yml
kevinwlau: Ah, I missed this line, I removed it.

cherylking (approved):

PR #706, opened 2022-01-31
Gen feat build file fix

kathrynkodama:
When a build file dependency change is detected, add an additional check to only call generateFeatures if there are class files to generate for.

Also include a minor improvement to the clean up logic in the dev integration tests. 

See https://github.com/OpenLiberty/ci.maven/pull/1409

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
@@ -566,20 +566,19 @@ class DevTask extends AbstractFeatureTask {
                 return true;
             } else if (installFeatures) {
                 if (generateFeatures) {
-                    // TODO confirm a call to generate features is required when a build file's compilation dependencies are modified
-                    // Increment generate features on build dependency change
+                    // TODO: if we generate features here we will also need to skip installing features on a failure
                     ProjectConnection gradleConnection = initGradleProjectConnection();
                     BuildLauncher gradleBuildLauncher = gradleConnection.newBuild();
                     runGradleTask(gradleBuildLauncher, 'compileJava', 'processResources'); // ensure class files exist
 
-                    // only call generateFeatures if classes
+                    // optimize generate features on build dependency change
                     Collection<String> javaSourceClassPaths = getJavaSourceClassPaths();
-                    if (!javaSourceClassPaths.isEmpty()) {
-                        if (libertyGenerateFeatures(javaSourceClassPaths, false)) {
-                            util.javaSourceClassPaths.clear();
-                        };
-                        libertyCreate(); // need to run create in order to copy generated config file to target
-                    }
+                    boolean generateFeaturesSuccess = libertyGenerateFeatures(javaSourceClassPaths, true);
+                    if (generateFeaturesSuccess) {
+                        util.javaSourceClassPaths.clear();
+                    };
+                    libertyCreate(); // need to run create in order to copy generated config file to target
+
                 }
                 libertyInstallFeature();
             }
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
@@ -566,13 +566,20 @@ class DevTask extends AbstractFeatureTask {
                 return true;
             } else if (installFeatures) {
                 if (generateFeatures) {
+                    // TODO confirm a call to generate features is required when a build file's compilation dependencies are modified
                     // Increment generate features on build dependency change
                     ProjectConnection gradleConnection = initGradleProjectConnection();
                     BuildLauncher gradleBuildLauncher = gradleConnection.newBuild();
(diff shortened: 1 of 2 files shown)


turkeylurkey (approved):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -566,13 +566,20 @@ class DevTask extends AbstractFeatureTask {
...
                     ProjectConnection gradleConnection = initGradleProjectConnection();
                     BuildLauncher gradleBuildLauncher = gradleConnection.newBuild();
                     runGradleTask(gradleBuildLauncher, 'compileJava', 'processResources'); // ensure class files exist
+
+                    // only call generateFeatures if classes
                     Collection<String> javaSourceClassPaths = getJavaSourceClassPaths();
-                    libertyGenerateFeatures(javaSourceClassPaths, false);
-                    libertyCreate(); // need to run create in order to copy generated config file to target
+                    if (!javaSourceClassPaths.isEmpty()) {
+                        if (libertyGenerateFeatures(javaSourceClassPaths, false)) {
```
TrevCraw, line 578: Same as https://github.com/OpenLiberty/ci.maven/pull/1409/files#r797880194

kathrynkodama:
Updated the logic to detect dependency changes in the build.gradle, and run `generateFeatures --optimize=true`. If generateFeatures is successful, run `libertyCreate` and `installFeature` as well.

Tried to add a new integration test to cover this case, but altering the build.gradle through the integration tests results in the following error
```
Could not parse build.gradle org.gradle.api.ProjectConfigurationException: A problem occurred configuring root project 'basic-dev-project1643909649043'.
```

Opened a separate issue: https://github.com/OpenLiberty/ci.gradle/issues/708

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -564,17 +580,21 @@ class DevTask extends AbstractFeatureTask {
...
-                    Collection<String> javaSourceClassPaths = getJavaSourceClassPaths();
-                    libertyGenerateFeatures(javaSourceClassPaths, false);
-                    libertyCreate(); // need to run create in order to copy generated config file to target
-                }
+            } else if (installFeatures && !compileProject) {
                 libertyInstallFeature();
+            } else if (compileProject && generateFeatures) { // generate features if compile dependencies have been modified
+                ProjectConnection gradleConnection = initGradleProjectConnection();
+                BuildLauncher gradleBuildLauncher = gradleConnection.newBuild();
+                runGradleTask(gradleBuildLauncher, 'compileJava', 'processResources'); // ensure class files exist
```
TrevCraw, line 588: For Maven, we are not compiling on compilation dependency change correct? It seems like this would be a new thing for Gradle. Should we keep it consistent between the two plugins?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -564,17 +580,21 @@ class DevTask extends AbstractFeatureTask {
...
+
+                // optimize generate features on build dependency change
+                Collection<String> javaSourceClassPaths = getJavaSourceClassPaths();
+                boolean generateFeaturesSuccess = libertyGenerateFeatures(javaSourceClassPaths, true);
```
TrevCraw, line 592: Just realized we do not need to pass the `javaSourceClassPaths` in this case since we are optimizing. Can pass `null`.
This would also apply to https://github.com/OpenLiberty/ci.maven/blob/c4038512547184e25699d16492e6db05976fe307/liberty-maven-plugin/src/main/java/io/openliberty/tools/maven/server/DevMojo.java#L915

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
kathrynkodama: Also updated to no longer call `create` and `installFeature` since that will be handled by the rest of the dev mode flow and to match the behaviour in ci.maven.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -564,19 +580,15 @@ class DevTask extends AbstractFeatureTask {
                 // - start server
                 util.restartServer();
                 return true;
+            } else if (compileDependenciesChanged && generateFeatures) { // generate features if compile dependencies have been modified
+                // optimize generate features on build dependency change
+                boolean generateFeaturesSuccess = libertyGenerateFeatures(null, true);
```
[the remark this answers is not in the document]
kathrynkodama: The .each method in Groovy is null safe, see https://stackoverflow.com/questions/5472795/why-dont-i-get-a-nullpointerexception-in-groovy-in-this-case

TrevCraw (approved):

PR #713, opened 2022-02-16
Add new tests regarding generate features.

turkeylurkey:
Direct stdout and stderr to separate files to avoid corruption.
Refactor DevTest to move utility methods to a base class.

Part of https://github.com/OpenLiberty/ci.maven/issues/1395


kathrynkodama (approved):
Looks good to me. There will probably be merge conflicts with https://github.com/OpenLiberty/ci.gradle/pull/712 so let me know which you think we should merge first.

PR #714, opened 2022-02-22
Fix configDirectory up to date check

### LINKED ISSUE #686: Custom config files are not deployed unless clean is run
If I add additional files to `src/main/liberty/config`, they are deployed when I run the `libertyStart` or `libertyCreate` tasks.

However, if I make changes to these files, they are not redeployed when I run `libertyStart` or `libertyCreate`, unless I first run `clean`.

If I add the `--info` option, I can see that gradle is skipping `libertyCreate` because "it is up-to-date".

mattbsox:
Will use default `configDirectory` value if `null` during `libertyCreate` up to date check.

Fixes #686


cherylking:
@mattbsox Seems like all the builds failed.

mattbsox:
Yeah, looks like I need to add something if the default location doesn't exist.

cherylking (approved):

PR #718, opened 2022-03-01
Add alternate output directory case for installFeature and package tasks

### LINKED ISSUE #693: Specify different outputDirectory for package task
When running the package task, specify a different outputDirectory to avoid the following error messages from the `server package` command in Liberty.

```
[ant:server] Packaging server defaultServer.
[ant:server] Server defaultServer is running.
[ant:server] Server defaultServer package failed. It must be stopped before it can be packaged.
```

See the code [here](https://github.com/OpenLiberty/ci.maven/blob/main/liberty-maven-plugin/src/main/java/io/openliberty/tools/maven/server/PackageServerMojo.java#L131) for similar code in the Liberty Maven plugin.

mattbsox:
Fixes #693

### DIFF
--- a/src/test/groovy/io/openliberty/tools/gradle/LibertyPackage_DefaultOutputDir_Test.groovy
+++ b/src/test/groovy/io/openliberty/tools/gradle/LibertyPackage_DefaultOutputDir_Test.groovy
@@ -55,6 +55,9 @@ class LibertyPackage_DefaultOutputDir_Test extends AbstractIntegrationTest{
            BuildResult result = runTasksResult(buildDir, 'libertyPackage')
 
            assertFalse("Running server message found in logs.", result.getOutput().contains('[ant:server] Server defaultServer package failed. It must be stopped before it can be packaged.'))
+
+           assert new File(buildDir, 'build/liberty-alt-output-dir').exists()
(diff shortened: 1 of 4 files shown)


src/test/groovy/io/openliberty/tools/gradle/LibertyPackage_DefaultOutputDir_Test.groovy
```diff
@@ -0,0 +1,62 @@
...
+    static File buildDir = new File(integTestDir, "/liberty-package-defaultOutputDir-test")
+    static File buildFilename = new File(resourceDir, "liberty-package-defaultOutputDir.gradle")
+
+    @BeforeClass
+    public static void setup() {
+        createDir(buildDir)
+        copyBuildFiles(buildFilename, buildDir)
+        copySettingsFile(resourceDir, buildDir)
+        try {
+            runTasks(buildDir, 'installLiberty', 'libertyStart')
```
cherylking, line 37: How is this test case confirming that the alt output dir was determined correctly and used?

mattbsox: Good point. I'll add something to check that the alternate output dir exists after package. Just looking for the original error message at the moment.

cherylking (approved):

PR #720, opened 2022-03-07
Generate features when features list has been modified, triggered from config file change

kathrynkodama:
Part of https://github.com/OpenLiberty/ci.maven/issues/1431 See https://github.com/OpenLiberty/ci.common/pull/346

Adds the `serverFeaturesModified()` method to check the configuration directory in source and returns true if features have been modified. Supported new `suppressLogs` boolean in `ServerFeatureSupport` so that the dev mode console is not flooded with duplicate info and warning messages when checking if server features have been modified.

Adds the corresponding dev mode integration tests.


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -691,6 +692,49 @@ class DevTask extends AbstractFeatureTask {
...
+            generatedFiles.add(BinaryScannerUtil.GENERATED_FEATURES_FILE_NAME);
+            File serverXmlFile = server.serverXmlFile; // can be null
+            Set<String> features = servUtil.getServerFeatures(configDirectory, serverXmlFile, new HashMap<String, File>(), generatedFiles);
+            // exclude generated features from the features list
+            Set<String> generatedFeatures = servUtil.getServerXmlFeatures(null,
+                    new File(configDirectory, BinaryScannerUtil.GENERATED_FEATURES_FILE_PATH), null, null);
+            servUtil.setSuppressLogs(false); // re-enable logs from ServerFeatureUtil
+
+            if (features != null) {
+                // remove generaetd features in the case that they have been manually added to the server config
```
turkeylurkey, line 710: Typo

turkeylurkey (approved):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -691,6 +692,49 @@ class DevTask extends AbstractFeatureTask {
             }
         }
 
+        @Override
+        public boolean serverFeaturesModified(File configDirectory) {
```
TrevCraw, line 696: Is the only difference between ci.gradle and ci.maven for this method the `serverXmlFile` handling? If so, it would be better to have this method in ci.common with `serverXmlFile` as an extra parameter (or using the `serverXmlFile` var directly in DevUtil).

kathrynkodama: I think the logic around getting the features should still be duplicated in ci.maven and ci.gradle as the `ServerFeatureUtil` object comes from `ServerFeatureSupport` and `AbstractFeatureTask` respectively. However, I agree we should move the common logic for comparing the two feature lists to ci.common.

PR #733, opened 2022-04-18
Display warning if installFeature fails when generateFeatures is enabled and Liberty feature deps are in the build file

kathrynkodama:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1471

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
@@ -656,7 +656,7 @@ class DevTask extends AbstractFeatureTask {
 
 
         @Override
-        public void installFeatures(File configFile, File serverDir) {
+        public void installFeatures(File configFile, File serverDir, boolean generateFeatures) {
             ServerFeatureUtil servUtil = getServerFeatureUtil(true);
             Set<String> features = servUtil.getServerFeatures(serverDir, libertyDirPropertyFiles);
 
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -698,7 +698,12 @@ class DevTask extends AbstractFeatureTask {
                 } catch (BuildException e) {
                     // stdout/stderr from the installFeature task is sent to the terminal
                     // only need to log the actual stacktrace when debugging
-                    logger.debug('Failed to install features from configuration file', e);
+                    logger.error('Failed to install features from configuration file', e);
+                    if (generateFeatures && !project.configurations.getByName('libertyFeature').dependencies.isEmpty()) {
```
TrevCraw, line 702: See https://github.com/OpenLiberty/ci.maven/pull/1482#discussion_r852208361

kathrynkodama: Done

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -699,7 +700,15 @@ class DevTask extends AbstractFeatureTask {
...
                     logger.error('Failed to install features from configuration file', e);
-                    if (generateFeatures && !project.configurations.getByName('libertyFeature').dependencies.isEmpty()) {
+
+                    boolean containsConflictMessage = false;
+                    if (e.getCause() != null && e.getCause().getCause() != null && e.getCause().getCause().getCause() != null) {
+                        if (e.getCause().getCause().getCause().getMessage().contains(InstallFeatureUtil.CONFLICT_MESSAGE)) {
+                            // PluginExecutionException from installFeature will be 3 layers deep
+                            containsConflictMessage = true;
+                        }
+                    }
```
TrevCraw, line 710: We only need to check for this conflict message if this check is true:
```
if (generateFeatures && !project.configurations.getByName('libertyFeature').dependencies.isEmpty())
```
Maybe put the conflict check inside of that if statement? Otherwise it will run every time, but it may not be necessary.

TrevCraw, line 710: We only need to check for this conflict message if this check is true:
```
if (generateFeatures && !project.configurations.getByName('libertyFeature').dependencies.isEmpty())
```
Maybe put the conflict check inside of that if statement? Otherwise it will run every time, but it may not be necessary.
If it ends up being messy, and readability is an issue, we could make the conflict check a separate method (containsConflictMessage boolean could be a method call)

TrevCraw (approved):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
kathrynkodama: Good idea, I've updated the logic

TrevCraw (approved):

turkeylurkey (approved):

PR #752, opened 2022-05-18
When generating features do not write the file if the features have not

turkeylurkey:
changed. Should be the same as LMP.

Signed-off-by: Paul Gooderham <turkeyonmarblerye@gmail.com>

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
@@ -176,25 +176,31 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
         def newServerXmlSrc = new File(server.configDirectory, GENERATED_FEATURES_FILE_PATH);
         try {
             if (missingLibertyFeatures.size() > 0) {
-                // Create specialized server.xml
-                ServerConfigXmlDocument configDocument = ServerConfigXmlDocument.newInstance();
-                configDocument.createComment(HEADER);
-                Element featureManagerElem = configDocument.createFeatureManager();
-                configDocument.createComment(featureManagerElem, GENERATED_FEATURES_COMMENT);
-                for (String missing : missingLibertyFeatures) {
-                    logger.debug(String.format("Adding missing feature %s to %s.", missing, GENERATED_FEATURES_FILE_PATH));
-                    configDocument.createFeature(missing);
+                Set<String> existingGeneratedFeatures = getGeneratedFeatures(servUtil, newServerXmlSrc);
+                if (!missingLibertyFeatures.equals(existingGeneratedFeatures)) {
+                    // Create specialized server.xml
+                    ServerConfigXmlDocument configDocument = ServerConfigXmlDocument.newInstance();
+                    configDocument.createComment(HEADER);
+                    Element featureManagerElem = configDocument.createFeatureManager();
+                    configDocument.createComment(featureManagerElem, GENERATED_FEATURES_COMMENT);
+                    for (String missing : missingLibertyFeatures) {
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
```diff
@@ -176,25 +176,31 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
...
+                        logger.debug(String.format("Adding missing feature %s to %s.", missing, GENERATED_FEATURES_FILE_PATH));
+                        configDocument.createFeature(missing);
+                    }
+                    // Generate log message before writing file as the file change event kicks off other dev mode actions
+                    logger.lifecycle("Generated the following features: " + missingLibertyFeatures);
+                    // use logger.lifecycle so that message appears without --info tag on
+                    configDocument.writeXMLDocument(newServerXmlSrc);
+                    logger.debug("Created file " + newServerXmlSrc);
+                    // Add a reference to this new file in existing server.xml.
+                    def serverXml = findConfigFile("server.xml", server.serverXmlFile);
```
kathrynkodama, line 196: I know this is existing behaviour, but will this reference still work if they have a custom server.xml file name configured in the build.gradle? It looks like it is searching in the source config folder?

src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
```diff
@@ -224,6 +230,17 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
         servUtil.setLowerCaseFeatures(true);
         return existingFeatures;
     }
+
+    // returns the features specified in the generated-features.xml file
+    private Set<String> getGeneratedFeatures(ServerFeatureUtil servUtil, File generatedFeaturesFile) {
+        servUtil.setLowerCaseFeatures(false);
+        Set<String> genFeatSet = new HashSet<String>();
+        servUtil.getServerXmlFeatures(genFeatSet, configDirectory,
```
kathrynkodama, line 238:
```suggestion
        servUtil.getServerXmlFeatures(genFeatSet, server.configDirectory,
```

kathrynkodama: As a future item, we could refactor this logic to BinaryScannerUtil or another util class in ci.common as it is very similar across ci.maven & ci.gradle. We have similar logic (but not identical) in DevUtil: https://github.com/OpenLiberty/ci.common/blob/4b19329fd1475ce925979f37144f9f5041062aaf/src/main/java/io/openliberty/tools/common/plugins/util/DevUtil.java#L5589-L5596

src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
kathrynkodama: Tried this out and if I specify in the build.gradle a serverXmlFile location outside of the config directory, the generate features comment is not written. ie.
```gradle
liberty {
	server {
		serverXmlFile = file("src/main/liberty/testing/server1.xml")
	}
}
```
I realize we are doing the same in ci.maven as well: https://github.com/OpenLiberty/ci.maven/blob/339fc5ff2b312b3a29dafa632fa0a998b14f96fb/liberty-maven-plugin/src/main/java/io/openliberty/tools/maven/server/GenerateFeaturesMojo.java#L255

Maybe we should open a new issue to address this.

turkeylurkey: Great, thank you.

kathrynkodama (approved):

Issue #781, opened 2022-09-28
GenerateFeaturesTest consistent failures on Windows due to file deletion issues


cherylking:
All the Windows builds are failing consistently and mostly due to GenerateFeaturesTest problems with deleting files. The DevTest also fails occasionally on Windows due to file deletion problems. Out of 15 jobs, only 6 passed. Of the 9 that failed, 7 were on Windows and 4 of those 7 failed only due to GenerateFeaturesTest.

Here's an example of the failures on a recent build of mine.

![Screen Shot 2022-09-28 at 2 25 28 PM](https://user-images.githubusercontent.com/29436451/192871166-380eabf5-4efc-44ae-9dc7-da26359ec4d5.png)

![Screen Shot 2022-09-28 at 2 26 00 PM](https://user-images.githubusercontent.com/29436451/192871189-fcd8449b-8bc9-40a0-a5b3-94c8f41fdcb0.png)

PR #784, opened 2022-10-20
Update common 1.8.23 and ci.ant 1.9.11

cherylking:
In addition to pulling in latest ci.ant and ci.common releases, made some yaml updates. Also I am excluding the GenerateFeaturesTest on Windows builds due to issue #781. When someone looks at that issue, they should reenable the test for Windows.


evie-lau (approved):
Changes look good. Note to keep on eye on the uploaded artifacts after builds are done

PR #797, opened 2023-02-20
Fix loose app generation for jar project dependencies in war projects

mattbsox:
Loose app xml files for war files will now include an archive element for jar project dependencies.

### DIFF
--- a/src/test/resources/loose-war-with-loose-jar/ejb-ejb/src/main/java/wasdev/ejb/ejb/SampleStatelessBean.java
+++ b/src/test/resources/loose-war-with-loose-jar/ejb-ejb/src/main/java/wasdev/ejb/ejb/SampleStatelessBean.java
@@ -1,0 +1,11 @@
+package wasdev.ejb.ejb;
+
+import javax.ejb.Stateless;
+
+@Stateless
+public class SampleStatelessBean {
+
+    public String hello() {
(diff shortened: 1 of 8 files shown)


src/test/resources/loose-war-with-loose-jar/ejb-ejb/src/main/java/wasdev/ejb/ejb/SampleStatelessBean.java
```diff
@@ -0,0 +1,26 @@
+/*******************************************************************************
```
cherylking, line 1: This copyright seems out of place. In general we don't have copyrights in test case files.

src/test/resources/loose-war-with-loose-jar/ejb-war/src/main/java/wasdev/ejb/ejb/web/EJBServlet.java
```diff
@@ -0,0 +1,49 @@
+/*******************************************************************************
```
cherylking, line 1: same comment about copyright

cherylking (approved):

cherylking:
I just noticed that all the failing builds are Java 8. Is there something to that? I didn't look at the build logs yet.

cherylking (approved):

PR #802, opened 2023-03-11
Dev mode debug

scottkurz:


scottkurz:
One thing I noticed in working on this is that sometimes the "port in-use detection" fails, i.e. the ci.common findAvailablePort thinks I can use a port which another JVM is already using as the debug port. 

I'm mostly running on Windows, which is the only place I've seen the problem (and haven't seen it on Linux).

@turkeylurkey not sure if you have any insight here.  I see there was a later round of work specifically on Windows:  https://github.com/OpenLiberty/ci.common/pull/230

Thanks

turkeylurkey:
I think @TrevCraw could help with this.

scottkurz:
Just to be clear... I don't think this PR is making whatever shortcomings may exist in 'findAvailablePort' any worse.. Or, to the extent it is worse, well, it's on a path (dev mode gradle debug) that was completely unusable before.

So this PR could be merged without tackling that question at all if we want.

turkeylurkey:
The reason I wanted Trevor to look is that I thought dev mode has a dependency on the preceding steps.

TrevCraw (approved):

PR #806, opened 2023-03-28
Additional support for multi-module loose ear applications

mattbsox:
The loose ear logic will now check for additional project dependencies inside of war project dependencies and handle them accordingly.

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
@@ -21,7 +21,8 @@ import io.openliberty.tools.common.plugins.config.ApplicationXmlDocument
 import io.openliberty.tools.common.plugins.config.LooseApplication
 import io.openliberty.tools.common.plugins.config.LooseConfigData
 import io.openliberty.tools.common.plugins.config.ServerConfigDocument
-import io.openliberty.tools.common.plugins.util.DevUtil;
+import io.openliberty.tools.common.plugins.util.DevUtil
+import io.openliberty.tools.common.plugins.util.OSUtil
 
 import org.apache.commons.io.FilenameUtils
 import org.gradle.api.artifacts.UnknownConfigurationException
@@ -42,6 +43,7 @@ import java.lang.NumberFormatException
 import java.nio.file.Files
 import java.nio.file.StandardCopyOption
 import java.text.MessageFormat
+import java.util.HashSet
 import java.util.regex.Matcher
 import java.util.regex.Pattern
 import java.io.File
@@ -345,7 +347,7 @@ class DeployTask extends AbstractServerTask {
         return false;
     }
 
-    private void addWarEmbeddedLib(Element parent, LooseWarApplication looseApp, Task task) throws Exception {
+    private void addWarEmbeddedLib(Element parent, LooseApplication looseApp, Task task) throws Exception {
         ArrayList<File> deps = new ArrayList<File>();
         task.classpath.each {
             deps.add(it)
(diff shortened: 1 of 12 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -522,9 +527,9 @@ class DeployTask extends AbstractServerTask {
 
     private void addEmbeddedLib(Element parent, Project project, LooseApplication looseApp, String dir) throws Exception {
         try {
-            if (project.configurations.getByName('compile') != null) {
+            if (project.configurations.findByName('impementation') != null) {
                 //Get only the compile dependencies that are included in the war
-                File[] filesAsDeps = project.configurations.compile.minus(project.configurations.providedCompile).getFiles().toArray()
+                File[] filesAsDeps = project.configurations.implementation.minus(project.configurations.providedCompile).getFiles().toArray()
```
cherylking, line 532: Fix the comment before this line. Also, are we sure we no longer have to look for compile dependencies (e.g. maybe when using older versions of Gradle)?

src/test/resources/multi-module-loose-ear-test/ejb-war/src/main/java/wasdev/ejb/ejb/web/EJBServlet.java
```diff
@@ -0,0 +1,49 @@
+/*******************************************************************************
+ * (c) Copyright IBM Corporation 2017.
```
cherylking, line 2: These copyrights in the test files seem wrong. I would remove them.

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -64,7 +64,7 @@ public class LooseEarApplication extends LooseApplication {
                     config.addFile(moduleArchive, f, "/WEB-INF/lib/" + f.getName());
                     break
                 case "MF":
-                    addManifestFileWithParent(moduleArchive, f)
+                    addManifestFileWithParent(moduleArchive, f, proj.sourceSets.main.getOutput().getResourcesDir().getParentFile().getCanonicalPath())
```
cherylking, line 67: How do we know this is the right path? A comment might help understand what path this is getting and why.

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
mattbsox: I can add a check for both, but the compile configuration will cause an error to be thrown with Gradle 7+.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -357,8 +359,12 @@ class DeployTask extends AbstractServerTask {
...
         for (File dep: deps) {
             String dependencyProjectName = getProjectPath(parentProjectDir, dep)
-            String projectDependencyString = "project ':" + dependencyProjectName + "'"
-            Project siblingProject = project.getRootProject().findProject(dependencyProjectName)
+            String projectDependencyString
+            Project siblingProject
+            if (dependencyProjectName != null && !dependencyProjectName.isEmpty()) {
+                projectDependencyString = "project ':" + dependencyProjectName + "'"
+                siblingProject = project.getRootProject().findProject(dependencyProjectName)
+            }
```
cherylking, line 367: Question: Now that `projectDependencyString` can be null on the next line, the `isCurrentProject` would equal false in that case. This would cause the code to go into the final `else` statement. Is that the expected behavior?

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -553,11 +570,21 @@ class DeployTask extends AbstractServerTask {
...
+            pattern = Pattern.compile("\\\\build\\\\.*")
+        } else {
+            pattern = Pattern.compile("/build/.*")
+        }
         Matcher matcher = pattern.matcher(dependencyPathPortion)
         projectPath = matcher.replaceAll("")
+        if (OSUtil.isWindows()) {
+            projectPath = projectPath.replace("\\", "")
+        } else {
+            projectPath = projectPath.replace("/", "")
```
cherylking, line 586: Why are the forward/backward slashes in the projectPath getting replaced with nothing? Is this just getting rid of a leading slash? If so, a substring might be better. If not, I don't understand how we can just get rid of all slashes in the projectPath.

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
mattbsox: Yeah, that was just for the leading slash. I can try with a substring instead.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
mattbsox: Yes, that is expected. There was an issue where the non project dependencies were getting handled as project dependencies so I added the project check.

cherylking (approved):

PR #814, opened 2023-04-28
Fix loose app tests for Windows

mattbsox:


cherylking (commented):
Can you go ahead and update the yaml file to include these tests on Windows?

src/test/groovy/io/openliberty/tools/gradle/TestLooseApplication.groovy
```diff
@@ -87,16 +89,24 @@ public class TestLooseApplication extends AbstractIntegrationTest{
...
       String nodeValue = nodes.item(0).getAttributes().getNamedItem("sourceOnDisk").getNodeValue();
-      assert nodeValue.endsWith("/commons-text-1.1.jar") && nodeValue.contains("/test-loose-application/build/libs/") : 'archive sourceOnDisk attribute value not correct'
 
+      if (OSUtil.isWindows()) {
+          Assert.assertTrue('archive sourceOnDisk attribute value not correct', nodeValue.endsWith("\\commons-text-1.1.jar") && nodeValue.contains("\\test-loose-application\\build\\libs\\"))
+      } else {
+          Assert.assertTrue('archive sourceOnDisk attribute value not correct', nodeValue.endsWith("/commons-text-1.1.jar") && nodeValue.contains("/test-loose-application/build/libs/")) 
+      }
 
       Assert.assertEquals("archive targetInArchive attribute value", "/WEB-INF/lib/commons-lang3-3.5.jar",
```
cherylking, line 99: Why doesn't this comparison need to be OS specific?

src/test/groovy/io/openliberty/tools/gradle/TestLooseApplicationWithWarTask.groovy
```diff
@@ -86,15 +88,23 @@ public class TestLooseApplicationWithWarTask extends AbstractIntegrationTest{
```
cherylking, line 88: Why doesn't this comparison at the above line need to be OS specific for `targetInArchive`?

src/test/groovy/io/openliberty/tools/gradle/TestLooseEarApplication.groovy
```diff
@@ -109,10 +111,13 @@ public class TestLooseEarApplication extends AbstractIntegrationTest{
           if (node.getAttributes().getNamedItem("targetInArchive").getNodeValue().equals("/WEB-INF/lib/commons-text-1.1.jar")) {
```
cherylking, line 111: same question

cherylking (changes requested):

src/test/groovy/io/openliberty/tools/gradle/TestLooseApplication.groovy
mattbsox: Seems that the targetInArchive values use the same formatting regardless of the OS. I'd guess that's just due to how the plugins build those paths rather than being something we get back from the OS/Gradle/Java like the other values we're checking that are OS dependent.

cherylking: Answering my own question....`targetInArchive` is relative to the contents of the archive...those always have forward slashes...it is expected.

cherylking (approved):

PR #815, opened 2023-05-01
Convert one test to use Kotlin

cherylking:
Converted one of our test cases to use Kotlin syntax for the build file (based on a customer trying to figure out the correct syntax for copying a dependency). This test case now has the copy dependency task added.


src/test/resources/server-config/testCreateLibertyConfigDir.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+        from(jdbcLib)
+        into("$buildDir/wlp/usr/shared/resources")
+        include("*.jar")
+    }
+
+    integrationTest {
+        dependsOn(libertyCreate)
+    }
+
+    libertyCreate {
```
sreich, line 55: hey, i'm curious...why is this using libertyCreate versus say, liberty { server { name = serverName ... } }, is that because it is happening at a different stage/phase or..?

i am guessing it's something to do with this being testing specific? just trying to get clarity between this and the docs

cherylking: It was just convenient to set the config parameters here since I needed to add the `dependsOn(copyJdbcLib)` to the `libertyCreate` task in order to get that task to run at the appropriate time. I was also having some trouble in the beginning getting the build to recognize `liberty { ...}` config because of how I was applying our plugin. Even after fixing that, I found that using `liberty { server {...}}` got an error though. I had to do the following instead:

```
liberty {
    server.name = serverName
    server.configDirectory = file("${project.projectDir}/src/test/resources")
}
```

mattbsox (approved):
Looks good!

mattbsox (approved):

src/test/resources/server-config/testCreateLibertyConfigDir.gradle.kts
sreich: okay, thanks for the explanation..i have a theory actually...i think this needs DelegateClosureOf<> to fix the issue you had. I ran into the same thing, but just worked around it for now

see https://docs.gradle.org/current/userguide/kotlin_dsl.html#groovy_closures_from_kotlin

```

liberty {

    server(delegateClosureOf<LibertyExtension> {
//actually this doesn't work as i expect, i can't access e.g. bootstrapProperties from within here, implicitly. unsure why or what proper fix is
    })

}

```

you can also use kotlin's builtin `apply` extension, which changes this scope within that closure. slightly less good than whatever the proper solution is. but, we should probably figure that out, too, and recommend that...

```

liberty {
server.apply {
//within this closure, server is 'this' scoped...
    name = serverName
    configDirectory = file("${project.projectDir}/src/test/resources")
}
}

```

cherylking: I'm going to mark this as resolved, since my test case is passing.

PR #824, opened 2023-05-17
Configure clean to depend on libertyStop

### LINKED ISSUE #822: Cleaning while executing libertyRun fails, then orphans the process with no way to shutdown
I've been encountering this quite often unintentionally...if you have libertyRun being executed, `gradle clean` will fail because some files are in use...So you run libertyStop, but then it can't stop the server because the clean task already got rid of what was needed...

Result: the build is both not cleaned, and the process is running in the background and has to get killed manually...

Expected: I'd expect `gradle clean` to realize that the server is running and fail & message with that. This should prevent it from cleaning files it needs and getting in a state it can't come back from

gradle plugin 3.5.2, gradle 7.x

cherylking:
Fixes #822


mattbsox (approved):
Looks good to me!

PR #854, opened 2023-10-19
Test on Java 21 with Liberty 23.0.0.10

mattbsox:


build.gradle
```diff
@@ -60,7 +60,7 @@ compileTestGroovy {
 }
 
 def libertyAntVersion = "1.9.13"
-def libertyCommonVersion = "1.8.29"
+def libertyCommonVersion = "1.8.30-SNAPSHOT"
```
cherylking, line 63: Shouldn't you be pulling in a new ci.ant version too? You made a change there for Java 21.

cherylking: same comment for LMP btw

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractFeatureTask.groovy
```diff
@@ -291,5 +293,17 @@ public class AbstractFeatureTask extends AbstractServerTask {
...
 
+    protected void setContainerEngine(AbstractContainerSupportUtil util) throws PluginExecutionException {
+        String LIBERTY_DEV_PODMAN = "liberty.dev.podman";
+        Map<String, Object> projectProperties = project.getProperties();
+        if (!projectProperties.isEmpty()) {
+            Object isPodman = projectProperties.get(LIBERTY_DEV_PODMAN);
+            isPodman = projectProperties.get(LIBERTY_DEV_PODMAN);
+            if (isPodman != null) {
+                util.setIsDocker(!(Boolean.parseBoolean(isPodman.toString())));
+                logger.debug("liberty.dev.podman was set to: " + (Boolean.parseBoolean(isPodman.toString())));
```
cherylking, line 304: Same comments as for LMP (why get the property twice, why not use containsKey, and check the string for a valid true/false and give warning before calling `setIsDocker`)

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -378,6 +380,8 @@ class DevTask extends AbstractFeatureTask {
             project.configurations.getByName('libertyFeature').dependencies.each {
                 dep -> this.existingLibertyFeatureDependencies.add(dep.name)
             }
+
+            setContainerEngine(this)
```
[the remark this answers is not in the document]
cherylking: I'm thinking of `InstallFeatureTaskUtil` which extends from `InstallFeatureUtil`. Seems like you should call the `setContainerEngine` from that constructor.

PR #880, opened 2024-02-16
Detect install changes in dev mode

### LINKED ISSUE #878: Failure to update liberty-plugin-config.xml during Dev Mode startup and after build file changes
Upon starting dev mode, the liberty-plugin-config.xml file does not reflect updates if the file pre-exists. Specifically, changes made to the installDir value within the build.gradle file are not being synchronised into the liberty-plugin-config.xml file. The same behaviour is observed when the installDir value is updated in the build.gradle while dev mode is running. The LMP works as expected in both scenarios, ie. the installDir value is updated in liberty-plugin-config.xml both at startup and while dev mode is running.

### Scenario 1

**Prerequisites:**
- The testing environment is the liberty.gradle.test.wrapper.app project, located at liberty-tools-vscode/src/test/resources/gradle.
- The project should already include a liberty-plugin-config.xml file.

**Steps to Reproduce:**
1. In the build.gradle file, incorporate a custom installDir value. For example, change the installDirectory by adding the following block:
`ext { 
     liberty.installDir = "/Users/dessina/Documents/Workspace/IntelliJ/guide-getting-started/start/myinstalldir/wlp" 
  } `
2. Start the liberty server.
3. Observe that the <installDirectory> value within the liberty-plugin-config.xml file remains unchanged, disregarding the new value specified in build.gradle.

### Scenario 2

**Prerequisites:**
Test environment is set up in the liberty.gradle.test.wrapper.app project, located under liberty-tools-vscode/src/test/resources/gradle.

**Steps to Reproduce:**

1. Perform a project clean operation using the command gradle clean.
2. Start liberty server with the default build.gradle file without modifying any configuration.
3. Modify the build.gradle file by adding a custom configuration. For example, change the installDirectory by adding the following block:
`ext { 
     liberty.installDir = "/Users/dessina/Documents/Workspace/IntelliJ/guide-getting-started/start/myinstalldir/wlp" 
  } `
4. Verify the <installDirectory> value in the liberty-plugin-config.xml file.
5. Observe that the value is not updated to the new path, indicating an issue with the live update functionality.

cherylking:
Fixes #878 (Scenario 2)

Detect changes in Liberty installation configuration during dev mode and throw error.

Need to do something similar in LMP. Will work on the PR next week.

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
@@ -538,6 +538,15 @@ class DevTask extends AbstractFeatureTask {
                 return false;
             }
 
+            // Detect change in installation configuration that requires restart of dev mode. Throw error.
+            if (hasInstallationConfigChanged(newProject, project)) {
+                // Note that a change in some config values requires a 'clean' because the install location is the same, but the
+                // artifact that gets installed would be different. This can happen when using 'libertyRuntime' for example and
+                // only changing the 'version'. The 'installLiberty' task cannot detect that difference today and would report the task as upToDate.
+                // It only detects changes in install location currently. It would not be trivial to detect the other types of changes.
+                throw new PluginExecutionException("A change in Liberty runtime installation configuration requires a 'clean'. After running the 'clean' task, please run the 'libertyDev' task again for the change to take effect.");
+            }
+
             if(hasServerConfigBootstrapPropertiesChanged(newProject, project)) {
                 logger.debug('Bootstrap properties changed');
                 restartServer = true;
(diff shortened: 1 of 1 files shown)


cherylking:
This is what the logging/error looks like:

Screenshot 2024-02-20 at 10 07 10 AM

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -688,6 +699,59 @@ class DevTask extends AbstractFeatureTask {
...
+            // The following project configuration is used for the Liberty installation
+            // liberty extension: installDir, baseDir, cacheDir, outputDir, userDir, runtime
+            // libertyRuntime dependency
+            // liberty->install block: type, runtimeUrl, version, useOpenLiberty
+            boolean foundChange = false
+            if (newProject.liberty.installDir != oldProject.liberty.installDir) {
+                logger.debug('liberty.installDir changed')
+                foundChange = true
+            }
+            if (newProject.liberty.baseDir != oldProject.liberty.baseDir) {
```
TrevCraw, line 712: A similar method on the LMP side used an `else-if` pattern and returned `true` as soon as a change was found. Should we follow that strategy here or do we want to make sure to debug every line that was changed?

cherylking: Right...I was trying to give as much feedback as possible if multiple things changed. But I could switch it to just return true on the first change detected. I don't have a problem with that. Let me know if you want me to change it.

cherylking: I changed it. Didn't see the point in an `else-if` pattern, but I did `return true` as soon as an `if` evaluates to `true`.

TrevCraw (approved):

PR #895, opened 2024-07-15
Update messages and docs related to running tests when skipTests is set to true

### LINKED ISSUE #889: Questions on skipTests behavior
According to docs: skipTests 	If this option is enabled, do not run any tests in dev mode. The default value is false.

What does behavior does this create exactly? 

* Does Dev mode by default run tests after every change? I thought that was only the hot tests command?

* How about for first run on dev mode?

* Does this effectively disable explicitly telling it to run tests, like when pressing the enter command?

cherylking:
Fixes #889


cherylking:
@dshimo I will remove the yaml changes for using my branch and skipping the VerifyFeatureTest before merging. Sarah fixed the failing test in a separate PR that I just merged.

dshimo (approved):

PR #904, opened 2024-09-09
removing use of deprecated plugin conventions

arunvenmany-ibm:
1. archivesBaseName in basepluginconvention is deprecated :- replace with baseextension base.archivesName
Refer:- https://discuss.gradle.org/t/android-defaultconfig-archivesbasename-deprecated/46384

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -178,7 +178,7 @@ pluginBundle {
 publishing {
     publications {
         libertyGradle (MavenPublication) {
-            artifactId = base.archivesName
+            artifactId = 'liberty-maven-plugin'
 
             from components.java
             artifact groovydocJar
--- a/build.gradle
+++ b/build.gradle
@@ -178,7 +178,7 @@ pluginBundle {
 publishing {
     publications {
         libertyGradle (MavenPublication) {
-            artifactId = 'liberty-maven-plugin'
+            artifactId = 'liberty-gradle-plugin'
 
             from components.java
             artifact groovydocJar
--- a/build.gradle
+++ b/build.gradle
@@ -1,17 +1,14 @@
-import java.io.FileOutputStream
-import java.io.IOException
-import java.io.OutputStream
-import java.util.Properties
 
 apply plugin: 'groovy'
 apply plugin: 'maven-publish'
 apply plugin: 'signing'
 apply plugin: "com.gradle.plugin-publish"
 
-archivesBaseName = 'liberty-gradle-plugin'
 group = 'io.openliberty.tools'
 version = '3.8.4-SNAPSHOT'
-
+base {
+    archivesName='liberty-gradle-plugin'
+}
 repositories {
     mavenLocal()
     mavenCentral()
@@ -116,29 +113,29 @@ test {
 
         try {
 
-        output = new FileOutputStream("${buildDir}/gradle.properties");
+            output = new FileOutputStream("${buildDir}/gradle.properties");
 
-        if (libertyRuntime == "ol") {
-            runtimeGroup = "io.openliberty"
-            runtimeArtifactId = "openliberty-runtime"
-            kernelArtifactId = "openliberty-kernel"
-        } else {
-            runtimeGroup = "com.ibm.websphere.appserver.runtime"
-            runtimeArtifactId = "wlp-javaee7"
-            kernelArtifactId = "wlp-kernel"
-        }
+            if (libertyRuntime == "ol") {
+                runtimeGroup = "io.openliberty"
+                runtimeArtifactId = "openliberty-runtime"
+                kernelArtifactId = "openliberty-kernel"
+            } else {
+                runtimeGroup = "com.ibm.websphere.appserver.runtime"
+                runtimeArtifactId = "wlp-javaee7"
+                kernelArtifactId = "wlp-kernel"
+            }
 
-        // set the properties value
-        prop.setProperty("lgpVersion", version)
-        prop.setProperty("runtimeGroup", runtimeGroup)
-        prop.setProperty("runtimeArtifactId", runtimeArtifactId)
-        prop.setProperty("kernelArtifactId", kernelArtifactId)
-        prop.setProperty("runtimeVersion", runtimeVersion)
-        prop.setProperty("antVersion", libertyAntVersion)
-        prop.setProperty("commonVersion", libertyCommonVersion)
+            // set the properties value
+            prop.setProperty("lgpVersion", version)
+            prop.setProperty("runtimeGroup", runtimeGroup)
+            prop.setProperty("runtimeArtifactId", runtimeArtifactId)
+            prop.setProperty("kernelArtifactId", kernelArtifactId)
+            prop.setProperty("runtimeVersion", runtimeVersion)
+            prop.setProperty("antVersion", libertyAntVersion)
+            prop.setProperty("commonVersion", libertyCommonVersion)
 
-        // save properties to project root folder
-        prop.store(output, null)
+            // save properties to project root folder
+            prop.store(output, null)
 
         } catch (IOException io) {
             io.printStackTrace()
@@ -181,11 +178,11 @@ pluginBundle {
 publishing {
     publications {
         libertyGradle (MavenPublication) {
-            artifactId = archivesBaseName
+            artifactId = base.archivesName
 
             from components.java
             artifact groovydocJar
-            artifact sourcesJar 
+            artifact sourcesJar
 
             pom {
                 name = 'liberty-gradle-plugin'
@@ -218,7 +215,7 @@ publishing {
 }
 
 if (project.hasProperty('ossrhUsername') && project.hasProperty('ossrhPassword')) {
-    
+
     if (!version.endsWith("SNAPSHOT")) {
         signing {
             sign publishing.publications.libertyGradle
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -1,17 +1,14 @@
...
 apply plugin: 'groovy'
 apply plugin: 'maven-publish'
 apply plugin: 'signing'
 apply plugin: "com.gradle.plugin-publish"
 
-archivesBaseName = 'liberty-gradle-plugin'
 group = 'io.openliberty.tools'
 version = '3.8.4-SNAPSHOT'
-
+base {
```
cherylking, line 9: Do we need to apply the `base` plugin? The example [here](https://docs.gradle.org/8.8/userguide/upgrading_version_8.html#base_convention_deprecation) shows configuring that.

build.gradle
```diff
@@ -181,11 +178,11 @@ pluginBundle {
 publishing {
     publications {
         libertyGradle (MavenPublication) {
-            artifactId = archivesBaseName
+            artifactId = 'liberty-gradle-plugin'
```
cherylking, line 181: Why aren't we using `base.archivesName` here?

src/main/groovy/io/openliberty/tools/gradle/tasks/CompileJSPTask.groovy
```diff
@@ -15,21 +15,11 @@
...
-import org.gradle.api.tasks.TaskAction
+import io.openliberty.tools.ant.jsp.CompileJSPs
+import org.apache.tools.ant.Project
 import org.gradle.api.Task
-import org.gradle.api.GradleException
+import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.bundling.War
-import org.gradle.api.logging.LogLevel
-
-import org.apache.tools.ant.Project;
```
cherylking, line 31: Why are you removing all of these imports? Seems like it would cause the Project for the `ant` variable to use the wrong Project (Gradle one instead).

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -1,42 +1,50 @@
...
         this.task = task
+        this.logger = logger
     }
 
     public void addSourceDir() throws Exception {
-        File sourceDir = new File(task.getProject().path.replace(":","") + "/" + task.getProject().getConvention().getPlugin(EarPluginConvention.class).getAppDirName())
-        config.addDir(sourceDir, "/")
+        if (task.getProject().getPlugins().hasPlugin("ear")) {
+            Ear ear = (Ear) task.getProject().ear
+            File sourceDir = new File(task.getProject().path.replace(":","") + "/" + ear.getAppDirectory().getAsFile().get().getAbsolutePath())
```
cherylking, line 26: What does `getAppDirectory()` return? Before we were getting just a directory name, not a full path. This seems like it is getting a full path, which when concatenated onto the project path would not be right.

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -1,42 +1,50 @@
...
+                config.addFile(applicationXmlFile, "/META-INF/application.xml");
+            } else {
+                applicationXmlFile = new File(task.getDestinationDirectory().get().getAsFile().getParentFile().getAbsolutePath() + "/tmp/ear" + applicationName);
+                config.addFile(applicationXmlFile, "/META-INF/application.xml");
+            }
         }
         else {
-            applicationXmlFile = new File(task.getDestinationDirectory().get().getAsFile().getParentFile().getAbsolutePath() + "/tmp/ear" + applicationName);
-            config.addFile(applicationXmlFile, "/META-INF/application.xml");
+            logger.warn("Project is expected to have EAR plugin. Application xml file may not be added correctly")
```
cherylking, line 47: "Could not get the application.xml file location from the EAR plugin because it is not configured. The file may not be added correctly to the application archive."

src/main/groovy/io/openliberty/tools/gradle/utils/LooseWarApplication.groovy
```diff
@@ -1,27 +1,34 @@
...
+
+        War war;
+        File sourceDir = new File("src/main/webapp")
+        if (task.getProject().getPlugins().hasPlugin("war")) {
+            war = (War) task.getProject().war
+            if (war.getWebAppDirectory().getAsFile().get() != null) {
+                sourceDir = new File(war.getWebAppDirectory().getAsFile().get().getAbsolutePath())
+            }
+        }else {
+            logger.warn("Default sourcedir path to src/main/webapp as WAR plugin does not exist. Application may not work expectedly")
```
cherylking, line 30: "Could not get the webAppDirectory location from the WAR plugin because it is not configured. Using the default location src/main/webapp instead. Application may not work as expected."

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -1,42 +1,50 @@
...
+            File applicationXmlFile = new File(task.getProject().path.replace(":", "") + "/" + ear.getAppDirectory().getAsFile().get().getAbsolutePath() + "/META-INF/" + applicationName)
+            if (applicationXmlFile.exists()) {
+                config.addFile(applicationXmlFile, "/META-INF/application.xml");
+            } else {
+                applicationXmlFile = new File(task.getDestinationDirectory().get().getAsFile().getParentFile().getAbsolutePath() + "/tmp/ear" + applicationName);
+                config.addFile(applicationXmlFile, "/META-INF/application.xml");
+            }
         }
         else {
-            applicationXmlFile = new File(task.getDestinationDirectory().get().getAsFile().getParentFile().getAbsolutePath() + "/tmp/ear" + applicationName);
```
cherylking, line 38: Why aren't we using a default in the else statement similar to before?

cherylking (changes requested):

build.gradle
cherylking: maybe if the `base` plugin is applied...it will work? See previous comment.

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
arunvenmany-ibm: Done

src/main/groovy/io/openliberty/tools/gradle/utils/LooseWarApplication.groovy
arunvenmany-ibm: Done

src/main/groovy/io/openliberty/tools/gradle/tasks/CompileJSPTask.groovy
cherylking: k..I see the ant one I was worried about is there. But having the IDE remove imports for Groovy might be risky. Because Groovy doesn't complain until you actually run the code (and have all paths exercised). I see List, Set, HashSet, File all referenced in this groovy file. Removing those imports doesn't make sense to me.

build.gradle
arunvenmany-ibm: its working even without adding base plugin explicitly. to avoid confusion, added apply plugin line for base plugin

build.gradle
arunvenmany-ibm: it works without and with apply base plugin. applied base plugin to resolve confusion

src/main/groovy/io/openliberty/tools/gradle/tasks/CompileJSPTask.groovy
arunvenmany-ibm: Build is successful after the reimports

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
arunvenmany-ibm: reordered code. please check now

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -1,42 +1,50 @@
...
+            applicationXmlFile = new File(task.getProject().path.replace(":", "") + "/" + ear.getAppDirectory().getAsFile().get().getAbsolutePath() + "/META-INF/" + applicationName)
+            if (applicationXmlFile.exists()) {
+                config.addFile(applicationXmlFile, "/META-INF/application.xml");
+            }
         }
-        else {
+        if (applicationXmlFile == null || !applicationXmlFile.exists()) {
             applicationXmlFile = new File(task.getDestinationDirectory().get().getAsFile().getParentFile().getAbsolutePath() + "/tmp/ear" + applicationName);
             config.addFile(applicationXmlFile, "/META-INF/application.xml");
+            logger.warn("Could not get the application.xml file location from the EAR plugin because it is not configured. The file may not be added correctly to the application archive.")
```
cherylking, line 47: Since we had default behavior before without a warning, I think we should not actually log a warning here. I didn't notice the default behavior since you had removed it previously.

cherylking (changes requested):
Just the one comment.

cherylking (approved):

PR #906, opened 2024-09-11
removing deprecated usage of project.builddir in all test projects

arunvenmany-ibm:


src/test/resources/arquillian-tests/skipWithXmlManaged/build.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
     }
 }
 
-project.buildDir = '../build'
+project.getLayout().setBuildDirectory('../build');
```
cherylking, line 16: should this really have the `;` at the end?

cherylking (changes requested):
Just the one question, but it applies to multiple gradle files.

src/test/resources/arquillian-tests/skipWithXmlManaged/build.gradle
arunvenmany-ibm: when using groovy, semicolon is optional. it works with or without semicolon

cherylking (approved):

PR #918, opened 2024-11-04
Gradle multi module 2.0

arunvenmany-ibm:


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -500,13 +499,23 @@ class DevTask extends AbstractFeatureTask {
         @Override
         public boolean updateArtifactPaths(ProjectModule projectModule, boolean redeployCheck, boolean generateFeatures, ThreadPoolExecutor executor)
                 throws PluginExecutionException {
-            // not supported for Gradle, only used for multi module Maven projects
+            // unable to identify the changes made, showing option for user. always return false because action is invoked manually
+            if (isMultiModuleProject()) {
```
cherylking, line 503: Where did this boolean get set?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -1431,40 +1454,15 @@ class DevTask extends AbstractFeatureTask {
...
-            for (MavenProject depProj : dependentProjects) {
-                dependentModules.add(depProj.getFile());
-            }
-            */
-            boolean upstreamSkipTests = false
-            boolean upstreamSkipITs = false
-            boolean upstreamSkipUTs = false
+            //get gradle project properties. It is observed that project properties contain all gradle properties
+            // properties are overridden automatically with the highest precedence
+            // in gradle, we are only using skopTests
```
cherylking, line 1459: skopTests should be skipTests

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -1431,40 +1454,15 @@ class DevTask extends AbstractFeatureTask {
...
+            boolean upstreamSkipTests = dependencyProject.hasProperty("skipTests") ? DevTaskHelper.parseBooleanIfDefined(dependencyProject.properties.get("skipTests")) : skipTests
 
             if (DevTaskHelper.getPackagingType(dependencyProject).equals("ear")) {
                 upstreamSkipUTs = true;
             }
-            // build list of dependent modules
+            // build list of dependent modules -> can be kept as empty list in ci.gradle as gradle itself is evaluating other project module dependencies
```
cherylking, line 1465: what?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -536,6 +545,13 @@ class DevTask extends AbstractFeatureTask {
             boolean installFeatures = false;
             boolean optimizeGenerateFeatures = false;
 
+            if (isMultiModuleProject()) {
+                info("We detected a change in build.gradle, but we cannot identify whether it's a runtime or dependency change");
+                info("r - to restart server, type 'r' and press Enter.");
+                info("Press Enter to recompile the project and run tests manually");
+                return false;
```
cherylking, line 552: "A change was detected in a build file. The libertyDev task could not determine if a server restart is required. To restart the server, type 'r' and press Enter."

I don't think we want to tell them about running tests here. If we do, it needs special logic to handle the fact that the key could have been changed from Enter to 't' plus Enter.

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
arunvenmany-ibm: Its set in DevUtil( already there as part of ci maven multi module changes. we are passing projectModuleList created by getProjectModules() as upstreamProjects)

image


image

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
arunvenmany-ibm: Message corrected

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
arunvenmany-ibm: Comment has been explained

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -500,13 +499,21 @@ class DevTask extends AbstractFeatureTask {
         @Override
         public boolean updateArtifactPaths(ProjectModule projectModule, boolean redeployCheck, boolean generateFeatures, ThreadPoolExecutor executor)
                 throws PluginExecutionException {
-            // not supported for Gradle, only used for multi module Maven projects
+            // unable to identify the changes made, showing option for user. always return false because action is invoked manually
+            if (isMultiModuleProject()) {
+                info("A change was detected in a build file. The libertyDev task could not determine if a server restart is required.");
+                info("To restart the server, type 'r' and press Enter.");
+            }
             return false;
```
cherylking, line 507: So by returning false in these two `updateArtifactPaths` nothing will be done (as your comment says). Should we return true instead so that the recompile is done? We still wouldn't know if a restart of the server was necessary, but at least we would take care of the case where it was a dependency change. And I think Gradle may optimize itself for the compile if there was no change that affected the compile (but I am not sure). What do you think?

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
arunvenmany-ibm: We can do that, we can set updateArtifacts to return true and recompileDependencies as true, this will trigger recompiling.. but due to the recompile related messages, the message we specified above will not be visible in the expected place. do you want to make the changes

cherylking: Ok, so what if we don't even call updateArtifactPaths if it is a multimodule project? In the two places updateArtifactPaths is called, we can check first if it is multimodule project. If it is, due the recompile and then print the message. What do you think?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -499,20 +499,20 @@ class DevTask extends AbstractFeatureTask {
         @Override
         public boolean updateArtifactPaths(ProjectModule projectModule, boolean redeployCheck, boolean generateFeatures, ThreadPoolExecutor executor)
                 throws PluginExecutionException {
-            // unable to identify the changes made, showing option for user. always return false because action is invoked manually
+            // for multi module, unable to identify the changes made, showing option for user. return true to trigger recompile
             if (isMultiModuleProject()) {
-                info("A change was detected in a build file. The libertyDev task could not determine if a server restart is required.");
-                info("To restart the server, type 'r' and press Enter.");
+                warn("A change was detected in a build file. The libertyDev task could not determine if a server restart is required.");
```
cherylking, line 504: Why did you take out the message about how to restart the server?

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -499,20 +499,20 @@ class DevTask extends AbstractFeatureTask {
...
         }
 
         @Override
         public boolean updateArtifactPaths(File parentBuildFile) {
-            // unable to identify the changes made, showing option for user. always return false because action is invoked manually
+            // for multi module, unable to identify the changes made, showing option for user. return true to trigger recompile
             if (isMultiModuleProject()) {
-                info("A change was detected in a build file. The libertyDev task could not determine if a server restart is required.");
-                info("To restart the server, type 'r' and press Enter.");
+                warn("A change was detected in a build file. The libertyDev task could not determine if a server restart is required.");
```
cherylking, line 514: same question

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -542,11 +542,10 @@ class DevTask extends AbstractFeatureTask {
             boolean restartServer = false;
             boolean installFeatures = false;
             boolean optimizeGenerateFeatures = false;
-
+            // for multi module, unable to identify the changes made, showing option for user. return true to trigger recompile
             if (isMultiModuleProject()) {
-                info("A change was detected in a build file. The libertyDev task could not determine if a server restart is required.");
-                info("To restart the server, type 'r' and press Enter.");
-                return false;
+                warn("A change was detected in a build file. The libertyDev task could not determine if a server restart is required.");
```
cherylking, line 547: same question

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
arunvenmany-ibm: As discussed earlier, there is a chance that due to the length of recompileJava build messages  our restart message may get written away.
What I understood from the previous comment was just to show the warning message. so the "Press r to Restart" part was removed

cherylking: Let's combine it into one long warning message then. We will get feedback at playback on changes after that.

cherylking (approved):

PR #919, opened 2024-11-07
For Var processing changes in ci.common testing 

arunvenmany-ibm:


src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -505,10 +505,9 @@ abstract class AbstractServerTask extends AbstractLibertyTask {
         configureMultipleAppsConfigDropins(serverNode)
     }
 
-    protected ServerConfigDocument getServerConfigDocument(CommonLogger log, File serverXML, File configDir, File bootstrapFile,
-            Map<String, String> bootstrapProp, File serverEnvFile, boolean giveConfigDirPrecedence, Map<String, File> libertyDirPropertyFiles) throws IOException {
-        if (scd == null || !scd.getServerXML().getCanonicalPath().equals(serverXML.getCanonicalPath())) {
-            scd = new ServerConfigDocument(log, serverXML, configDir, bootstrapFile, bootstrapProp, serverEnvFile, giveConfigDirPrecedence, libertyDirPropertyFiles)
+    protected ServerConfigDocument getServerConfigDocument(CommonLogger log, Map<String, File> libertyDirPropertyFiles) throws IOException {
+        if (scd == null) {
```
cherylking, line 509: So this seems problematic. Before we were checking to see if the path for the server.xml file changed and if so we recreated the ServerConfigDocument. Now we cannot tell if it changed. Makes me wonder if we should always create a new ServerConfigDocument, but that is a lot of processing that happens.

src/main/groovy/io/openliberty/tools/gradle/tasks/StartTask.groovy
```diff
@@ -86,8 +86,7 @@ class StartTask extends AbstractServerTask {
         if (serverConfigFile != null && serverConfigFile.exists()) {
             try {
                 Map<String,String> props = combinedBootstrapProperties == null ? convertPropertiesToMap(server.bootstrapProperties) : combinedBootstrapProperties;
-                getServerConfigDocument(new CommonLogger(project), serverConfigFile, server.configDirectory, server.bootstrapPropertiesFile, props, server.serverEnvFile, 
-                                                                            false, getLibertyDirectoryPropertyFiles(null));
+                scd = getServerConfigDocument(new CommonLogger(project), getLibertyDirectoryPropertyFiles(null));
```
cherylking, line 89: I don't think the line before this needs to be done anymore. The bootstrap.properties file in the server dir will be used and should have all the combined properties already.

src/main/groovy/io/openliberty/tools/gradle/tasks/UndeployTask.groovy
```diff
@@ -79,8 +79,7 @@ class UndeployTask extends AbstractServerTask {
             File serverXML = new File(getServerDir(project).getCanonicalPath(), "server.xml")
 
             try {
-                getServerConfigDocument(new CommonLogger(project), serverXML, server.configDirectory,
-                        server.bootstrapPropertiesFile, combinedBootstrapProperties, server.serverEnvFile, false, getLibertyDirectoryPropertyFiles(null))
+                scd= getServerConfigDocument(new CommonLogger(project), getLibertyDirectoryPropertyFiles(null))
```
cherylking, line 82: I don't think you need to get serverXML at line 79 any more.

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -682,8 +682,7 @@ class DeployTask extends AbstractServerTask {
                 File serverXML = new File(getServerDir(project).getCanonicalPath(), "server.xml")
 
                 try {
-                    scd = getServerConfigDocument(new CommonLogger(project), serverXML, server.configDirectory,
-                            server.bootstrapPropertiesFile, combinedBootstrapProperties, server.serverEnvFile, false, getLibertyDirectoryPropertyFiles(null))
+                    scd = getServerConfigDocument(new CommonLogger(project), getLibertyDirectoryPropertyFiles(null))
```
cherylking, line 685: You don't need to get serverXML anymore at line 682.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -520,8 +519,7 @@ abstract class AbstractServerTask extends AbstractLibertyTask {
         if (serverConfigFile != null && serverConfigFile.exists()) {
             try {
                 Map<String,String> props = combinedBootstrapProperties == null ? convertPropertiesToMap(server.bootstrapProperties) : combinedBootstrapProperties;
-                getServerConfigDocument(new CommonLogger(project), serverConfigFile, server.configDirectory, server.bootstrapPropertiesFile, props, server.serverEnvFile, 
-                                                                            false, getLibertyDirectoryPropertyFiles(null));
+                scd = getServerConfigDocument(new CommonLogger(project), getLibertyDirectoryPropertyFiles(null));
```
cherylking, line 522: Don't need to get props anymore at line 521.

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/UndeployTask.groovy
arunvenmany-ibm: Added back logic in getServerDocument() to compare server.xml

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
arunvenmany-ibm: Added back logic in getServerDocument() to compare server.xml

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
arunvenmany-ibm: added this logic back. I think we should keep the logic

PR #921, opened 2024-11-15
Making change to create thin application with same name as provided i…

arunvenmany-ibm:
…n server.xml <springBootApplication/> node


cherylking (changes requested):
Need to add two negative test scenarios:

1. More than one springBootApplication element specified to trigger the exception. Use runTasksFailResult to check for error in output.
2. Specify the apps folder but do not provide the configuration for the springBootApplication and test that the config gets generated in the configDropins folder.

src/test/groovy/io/openliberty/tools/gradle/TestSpringBootApplication30.groovy
```diff
@@ -49,6 +52,8 @@ public class TestSpringBootApplication30 extends AbstractIntegrationTest{
             Assert.assertEquals("Did not get expected http response.","Hello!", webPage)
             Assert.assertTrue('defaultServer/dropins has app deployed',
                     new File(buildDir, 'build/wlp/usr/servers/defaultServer/dropins').list().size() == 0)
+            Assert.assertTrue('defaultServer/configDropins/defaults has no config',
+                    new File(buildDir, 'build/wlp/usr/servers/defaultServer/configDropins/defaults').list().size() == 1)
```
cherylking, line 56: Can we verify the contents of the generated file?

cherylking (changes requested):

cherylking (approved):

PR #927, opened 2024-12-23
 multiple server restart issue fix for maven

arunvenmany-ibm:
adding constructor parameter with emptymap for ci.gradle


arunvenmany-ibm:
Application multi restart is reported in Maven, Gradle has build state check, which will not trigger recompile, unless there is any file change.
We are passing flag as "true" to trigger the recompile logic always, and let gradle handle the dirty check

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -387,6 +387,7 @@ class DevTask extends AbstractFeatureTask {
                     true /* useBuildRecompile */, true /* gradle */, pollingTest, container, containerfile, containerBuildContext, containerRunOpts, containerBuildTimeout, skipDefaultPorts,
                     null /* compileOptions not needed since useBuildRecompile is true */, keepTempContainerfile, mavenCacheLocation, projectModuleList /* multi module upstream projects */,
                     projectModuleList.size() > 0 /* recompileDependencies as true for multi module */, packagingType, buildFile, parentBuildGradle /* parent build files */, generateFeatures, null /* compileArtifactPaths */, null /* testArtifactPaths */, webResourceDirs /* webResources */
+                    ,Collections.emptyMap() /*ignored for gradle*/
```
cherylking, line 390: This PR would be unnecessary if you provided a constructor that did not take this new Map since it is not needed for Gradle. That constructor could then just turn around and call the other one passing an empty Map, but in ci.common instead of ci.gradle.

cherylking (changes requested):

arunvenmany-ibm:
Closing PR after adding a new constructor in ci.common DevUtil() for ci.maven, hence common changes will not affect ci.gradle

PR #930, opened 2025-01-22
gradle deprecation fixes testing

arunvenmany-ibm:
Creating a PR to test builds with --warning-mode=all and fixing issues


src/main/groovy/io/openliberty/tools/gradle/tasks/CompileJSPTask.groovy
```diff
@@ -118,10 +118,10 @@ class CompileJSPTask extends AbstractFeatureTask {
             } 
         }
         
-        if (project.hasProperty('sourceCompatibility')) {
+        if (project.getExtensions().getByName('java').hasProperty("sourceCompatibility")) {
             logger.info("Found sourceCompatibility")
-            compile.setSource((String) project.getProperties().get('sourceCompatibility'))
-        }
+            compile.setSource((String) project.getExtensions().getByName('java').getProperties().get('sourceCompatibility'))
```
cherylking, line 123: Can we be sure `getByName` won't return null? It may be fine, but I wasn't sure.

src/test/resources/sample.servlet/testLooseApplication.gradle
```diff
@@ -22,8 +22,10 @@ buildscript {
 apply plugin: 'war'
 apply plugin: 'liberty'
 
-sourceCompatibility = 1.8
-targetCompatibility = 1.8
+java {
```
cherylking, line 25: I just had a thought. What about our customers who want to still use Gradle 8.x? Are we forcing them to change how they specify the source/target compatibility? Do we need to make sure our code still works even if they have it defined outside the `java` extension block?

cherylking: In other words, maybe we need to tolerate both for now?

cherylking (changes requested):

src/test/resources/sample.servlet/testLooseApplication.gradle
arunvenmany-ibm: I am not that is possible. If we keep the code to support older edition(project conventions), we will get deprecation warning.

src/main/groovy/io/openliberty/tools/gradle/tasks/CompileJSPTask.groovy
arunvenmany-ibm: As per the documentation, if we have a java plugin used , extension will be automatically created by JavaBasePlugin class
https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.plugins/-java-base-plugin/index.html

src/test/resources/sample.servlet/testLooseApplication.gradle
cherylking: My real question is "Does the new code work with the old way of specifying source/targetCompatibility?" If not, then I think we have to defer this change because we will break many existing customers. We would have to do this with a major version change of our plugin.

arunvenmany-ibm: Okay. Then I think gradle 9 deprecation fixes should be released as a major release later
Main affected areas would be usage of project extensions where originally conventions were used. with this standard, main issue would be with JavaExtension

cherylking:
> Do we need to keep supporting spring boot 2, or should we mention it in release notes that "from gradle 9, spring boot gradle plugin support is removed for version 2.7.18 or below"

@arunvenmany-ibm As discussed in scrum, we should wait until Gradle 9 has been released and then see if the springboot plugin gets updated to support Gradle 9. We should keep supporting springboot 2 if possible.

cherylking (approved):
Just noting here that we decided any breaking changes (like the change to use the Java extension to get the source/targetCompatibility configuration) would need to be made in a separate PR and would require a bump in the major version of our plugin. This will be deferred until Gradle 9 is actually released and we can test with it and see if we can support both ways to configure those properties.

PR #932, opened 2025-02-12
changes to support variable resolution for predefined variables

arunvenmany-ibm:


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -1247,6 +1247,7 @@ class DevTask extends AbstractFeatureTask {
         File configDirectory = server.configDirectory;
         // getOutputDir returns a string
         File serverOutputDir = new File(getOutputDir(project));
+        File serverOutputDirectory= new File(serverOutputDir, serverName)
```
cherylking, line 1250: This is named too similarly to the variable before it. Probably the one before it should just be `outputDir`.

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
arunvenmany-ibm: Done

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -1292,7 +1292,7 @@ class DevTask extends AbstractFeatureTask {
                 verifyAppStartTimeout.intValue(), verifyAppStartTimeout.intValue(), compileWait.doubleValue(),
                 libertyDebug.booleanValue(), pollingTest.booleanValue(), container.booleanValue(), containerfile, containerBuildContext, containerRunOpts,
                 containerBuildTimeout, skipDefaultPorts.booleanValue(), keepTempContainerfile.booleanValue(), localMavenRepoForFeatureUtility,
-                DevTaskHelper.getPackagingType(project), buildFile, generateFeatures.booleanValue(), webResourceDirs, projectModules, parentBuildGradle
+                DevTaskHelper.getPackagingType(project), buildFile, generateFeatures.booleanValue(), webResourceDirs, projectModules, parentBuildGradle, new File(serverOutputDir, serverName)
```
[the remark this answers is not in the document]
arunvenmany-ibm: serverOutputDir here is the response of getOutputDir() from AbstractServerTask

image
if you see ServerStatusUtil.isServerRunning() method, its always appending serverName to the new File()
image

By this, I got the understanding that getOutputDir() returns a directory without serverName

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -1149,8 +1149,9 @@ abstract class AbstractServerTask extends AbstractLibertyTask {
         }
         File wlpInstallDir = getInstallDir(project)
         File wlpUserDir = getUserDir(project, wlpInstallDir)
+        File serverOutputDir= new File(getOutputDir(project), server.name)
```
cherylking, line 1152: This line is why I said that `serverOutputDir` in AbstractServerTask already had server.name added to it. Is this not the same variable you are using in DevTask at line 1295?

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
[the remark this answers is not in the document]
cherylking: The confusion comes from the variable declared at line 1249. It is not the server output directory...it is just the output directory....when you add the server name to it...then it is the serverOutputDirectory. Combine that with the declaration of the same named variable in AbstractServerTask (and it includes the server name), it is very confusing. Please change the variable at line 1249 to just `outputDirectory`.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
cherylking: Looks like it is not the same variable. It just added to the confusion.

arunvenmany-ibm: the serverOutputDir in AbstractServerTask is just a local variable. the serverOutputDir in DevTask is an existing code, which caused the confusion.  I have modified the existing variable to outputDir to provide more clear context

cherylking (approved):

Issue #938, opened 2025-03-09
How to get the bare minimum task executor for bootstrap.properties generation


sreich:
Hello, I was using the 'deploy' task which was generating my bootstrap.properties

I was wondering if there was a way to do the bare minimum -- i only need that file, and so the CI generating an entire feature set, well that can add like a minute of time just to do nothing beyond that file

Is there a lesser task I can do? I have tried 'build libertyCreate copyJdbcLib", but..it still seems to invoke install features.

cherylking:
@sreich There is not a way to create a liberty server without the features getting installed, and the server must be created for the config files such as bootstrap.properties to get created in the created server. The only shortcut I can give you is after the initial server creation, there is a `skipInstallFeature` [parameter](https://github.com/OpenLiberty/ci.gradle/blob/main/docs/libertyDev.md#command-line-parameters) available on `libertyDev` task.

sreich:
Ooh that does seem helpful, thanks.

I see this is not a property but a parameter. Similar to the changeOnDemandTestsAction...

Can you make this a property for me so I can expose it within our toolchain? Having to pass this to each task run, across the fleet of apps to dev, would be bothersome

PR #941, opened 2025-03-19
Add prop skip install feature

cherylking:
Related to https://github.com/OpenLiberty/ci.gradle/issues/938#issuecomment-2715731297.

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/extensions/DevExtension.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/extensions/DevExtension.groovy
@@ -53,6 +53,11 @@ class DevExtension {
     // If set to `true`, change the action for running on demand tests from `Enter` to type `t` and press `Enter`. The default value is `false`.
     boolean changeOnDemandTestsAction = false
 
+    // If set to `true`, the `installFeature` task will be skipped when `dev` mode is started on an already existing Liberty runtime installation. 
+    // It will also be skipped when `dev` mode is running and a restart of the server is triggered either directly by the user or by application changes. 
+    // The `installFeature` task will be invoked though when `dev` mode is running and a change to the configured features is detected. The default value is `false`.
+    boolean skipInstallFeature = false
+
     //Docker aliases to maintain backwards compatability
     File dockerfile
     File dockerBuildContext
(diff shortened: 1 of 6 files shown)


src/main/groovy/io/openliberty/tools/gradle/extensions/DevExtension.groovy
```diff
@@ -53,6 +53,11 @@ class DevExtension {
     // If set to `true`, change the action for running on demand tests from `Enter` to type `t` and press `Enter`. The default value is `false`.
     boolean changeOnDemandTestsAction = false
 
+    // If set to `true`, the `installFeature` task will be skipped when `dev` mode is started on an already existing Liberty runtime installation. 
+    // It will also be skipped when `dev` mode is running and a restart of the server is triggered either directly by the user or by application changes. 
+    // The `installFeature` task will be invoked though when `dev` mode is running and a change to the configured features is detected. The default value is `false`.
+    boolean skipInstallFeature = false
```
arunvenmany-ibm, line 59: can we set the default value using DEFAULT_SKIP_INSTALL_FEATURE

src/test/groovy/io/openliberty/tools/gradle/DevSkipInstallFeatureConfigTest.groovy
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.AfterClass;
+import org.junit.Test;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+class DevSkipInstallFeatureConfigTest extends BaseDevTest {
```
arunvenmany-ibm, line 32: Just asking, do you think we should add a test with skipInstallFeature as parameter and gradle property, to make sure the right precedence is considered in our code?

src/main/groovy/io/openliberty/tools/gradle/extensions/DevExtension.groovy
cherylking: I don't see any of the other Extension classes importing from other files and referencing constants for any fields. So I'd rather keep the same pattern here.

src/test/groovy/io/openliberty/tools/gradle/DevSkipInstallFeatureConfigTest.groovy
cherylking: I'll give it a try.

cherylking: I added a config setting to the existing test case to prove the cmd line option overrides the config setting.

arunvenmany-ibm (approved):

PR #951, opened 2025-05-22
Print an error message if generate features cannot find the server

turkeylurkey:
Related to #1879

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
@@ -71,6 +71,10 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
 
     @TaskAction
     void generateFeatures() {
+        if (!getServerDir(project).exists()) {
+            throw new GradleException("GenerateFeatures task requires the Liberty server but unable to find server path " + getServerDir(project) + ". Please execute the task libertyCreate.");
+        }
+
         binaryScanner = getBinaryScannerJarFromRepository();
         BinaryScannerHandler binaryScannerHandler = new BinaryScannerHandler(binaryScanner);
 


src/main/groovy/io/openliberty/tools/gradle/tasks/GenerateFeaturesTask.groovy
```diff
@@ -71,6 +71,10 @@ class GenerateFeaturesTask extends AbstractFeatureTask {
 
     @TaskAction
     void generateFeatures() {
+        if (!getServerDir(project).exists()) {
+            throw new GradleException("GenerateFeatures task requires the Liberty server but unable to find server path " + getServerDir(project) + ". Please execute the task libertyCreate.");
+        }
+
```
TrevCraw, line 77: Suggestion to follow the behaviour for Gradle installFeature task, where the dependent tasks will be invoked (i.e. create).

turkeylurkey: Added the depends on configuration. Note that createLiberty specifies finalized by installFeature so it runs after create. The problem is that generate runs after install. We must specify `finalizedBy 'installFeature'` on generate features because Gradle will only run installFeature once. This way it runs after generate which is more natural than before. This is not part of the initial specification of this task but it may make life easier for the user.

TrevCraw: I think this is good. Just to clarify behaviour, does this mean that when a user runs the generate-features Gradle task on its own that the create task will run before and the install task will run afterwards every time?

turkeylurkey: I believe so. It doesn't print anything when it runs create but it does for install. Doing a clean and then generate features it creates the server and afterwards it installs features with a message. Doing it again without a clean it is not clear if create does anything but it does print messages for generate and for install features.

TrevCraw: Thanks @turkeylurkey , I think we should update our docs with this info. Do we have an issue open for updating the feature gen docs?

turkeylurkey: Opened #1896

PR #1002, opened 2025-09-09
using compileClasspath and runtimeClasspath instead of implementation and compile

arunvenmany-ibm:
…

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
@@ -558,13 +558,15 @@ class DeployTask extends AbstractServerTask {
         try {
             Set<File> filesAsDeps = new HashSet<File>()
             //Get the compile and implementation dependencies that are included in the war
-            if (project.configurations.findByName('compile') != null) {
-                Set<File> compileDepFiles = project.configurations.compile.minus(project.configurations.providedCompile).getFiles()
-                filesAsDeps.addAll(compileDepFiles)
+            //compileClasspath: Use this to get all the dependencies required to compile your source code. It includes api, implementation, and compileOnly dependencies.
+            if (project.configurations.findByName('compileClasspath') != null) {
+                Set<File> compileClasspathDepFiles = project.configurations.compileClasspath.getFiles()
+                filesAsDeps.addAll(compileClasspathDepFiles)
             }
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -558,13 +558,15 @@ class DeployTask extends AbstractServerTask {
         try {
             Set<File> filesAsDeps = new HashSet<File>()
             //Get the compile and implementation dependencies that are included in the war
-            if (project.configurations.findByName('compile') != null) {
-                Set<File> compileDepFiles = project.configurations.compile.minus(project.configurations.providedCompile).getFiles()
-                filesAsDeps.addAll(compileDepFiles)
+            //compileClasspath: Use this to get all the dependencies required to compile your source code. It includes api, implementation, and compileOnly dependencies.
```
cherylking, line 561: There is one other place that we call `findByName('compile')` in AbstractServerTask. Should that change to use `compileClasspath`?

src/main/groovy/io/openliberty/tools/gradle/tasks/DeployTask.groovy
```diff
@@ -558,13 +558,15 @@ class DeployTask extends AbstractServerTask {
...
+            if (project.configurations.findByName('compileClasspath') != null) {
+                Set<File> compileClasspathDepFiles = project.configurations.compileClasspath.getFiles()
+                filesAsDeps.addAll(compileClasspathDepFiles)
             }
-            if (project.configurations.findByName('impementation') != null) {
-                Set<File> implementationDepFiles = project.configurations.implementation.minus(project.configurations.providedCompile).getFiles()
-                filesAsDeps.addAll(implementationDepFiles)
+            //runtimeClasspath: Use this to get all the dependencies required to run your application at runtime. It includes both implementation and runtimeOnly dependencies.
+                if (project.configurations.findByName('runtimeClasspath') != null) {
+                Set<File> runtimeClasspathDepFiles = project.configurations.runtimeClasspath.getFiles()
```
cherylking, line 568: Do either of these include `providedCompile` in the list of files (since there was special logic to remove them before this change)? It seems like it would since it is still including `implementation` dependencies.

cherylking (changes requested):

PR #1073, opened 2026-03-24
Loose EAR Manifest Class-Path Issue Fix

### LINKED ISSUE #766: Multi module  Gradle application failed to deploy into liberty server
I am trying to deploy multi module Gradle application into Liberty server and it is failing to deploy into apps folder.
I followed this documentation and I could not deploy into liberty server.

```
apply plugin: 'liberty'

dependencies {
    libertyApp 'example:app:1.0'
}
```

I tried both  looseApplication=false and true options and it did not work. Here is my gradle.build file below.

```
apply plugin: 'liberty'

buildscript {
    repositories {
        mavenCentral()
        maven {
            name = 'Sonatype Nexus Snapshots'
            url = 'https://oss.sonatype.org/content/repositories/snapshots/'
        }
    }
    dependencies {
        classpath 'io.openliberty.tools:liberty-gradle-plugin:3.3.1'
    }
}

description = 'GarageSaleRuntime'
def libertyInstallDir = "C:\\MyData\\LibertyInstalls\\wlp"

dependencies {
  libertyApp project(':GarageSaleLibertyEAR')
  libertyRuntime group: 'com.ibm.websphere.appserver.runtime', name: 'wlp-jakartaee9', version: '22.0.0.5'
}

liberty {
    installDir="${libertyInstallDir}"
    server {
        name = "garageSaleServer"
        serverXmlFile = file('publish/servers/server.xml')
        jvmOptionsFile = file('publish/files/jvm.options')
        configDirectory = file('publish/config')
        looseApplication = false
        stripVersion = true
        stripVersion = true
        	features {
            	name = ["${projectDir}/publish/files/WASPersonaWebServicesHandlerFeature_1.0.0.202110181027.esa"]
            	acceptLicense = true
        	}
    }
}

task copyDB2Jars(type: Copy) {
    from 'publish/databaseDrivers/db2V9Drivers'
    into "${buildDir}/wlp/usr/shared/resources/db2drivers"
} 

task copyJaxRSThirdPartyJars(type: Copy) {
    from 'publish/jaxrsThirdPartyJars'
    into "${buildDir}/wlp/usr/shared/resources/jaxrsThirdPartyJars"
}

task copyMyEAR(type: Copy) {
	from '../GarageSaleLibertyEAR/build/libs/GarageSaleLibertyEAR.ear'
	into "${buildDir}/wlp/usr/servers/garageSaleServer/apps"
}

//libertyCreate.finalizedBy 'copyMyEAR', 'copyDB2Jars', 'copyJaxRSThirdPartyJars'
libertyCreate.finalizedBy 'copyDB2Jars', 'copyJaxRSThirdPartyJars'
build.dependsOn 'deploy'

```

sajeerzeji:
Fixes #766 

### Issue
The application failed to start in loose EAR mode with,
```
StateChangeException: Error while loading class com.ibm.websphere.svt.gs.gsdb.session.MfgCategorySessionBean
ResourceLoadingException: Error while loading JPA entity classes (MfgCategory, Category, etc.)
```

In packaged EAR mode, JAR files are physically present in the EAR, and Liberty's classloader uses the manifest Class-Path entries to locate dependency JARs within the EAR.

In loose EAR mode (looseApplication = true), there are no physical JAR files. Instead, the plugin generates an XML descriptor (GarageSaleLibertyEAR.ear.xml) that maps source directories. Liberty's classloader needs to know which class directories belong to each module.

**The Problem**
- EJB modules depend on other modules (e.g., JPA modules) via `compileOnly` or `implementation` dependencies
- The plugin only added the EJB module's own classes to the loose XML
- Dependency class directories were not included, causing ClassNotFoundException at runtime

**Previous Approach (Reverted)**
The previous fix attempted to use manifest Class-Path entries to determine dependencies, requiring users to manually declare Class-Path in their build.gradle files. This approach was reverted after cross-checking with Maven plugin implementation.

**Current Approach**
Added `addDependencyClassDirectories()` method that,
1. Scans both `compileClasspath` and `runtimeClasspath` for project dependencies
2. Automatically adds each dependency's class and resource directories to the module in loose XML
3. Works without requiring manifest Class-Path configuration

**How it works**
- Plugin automatically detects project dependencies from Gradle configurations
- For each EJB/JAR module, it finds all project dependencies
- Adds dependency class directories as `<dir>` elements in the loose XML
- Liberty's classloader can now find all required classes at runtime

**Usage Requirements**
No special configuration needed. The plugin automatically handles dependencies declared in build.gradle.

**Backward Compatibility**
- Packaged EAR mode (`looseApplication = false`) continues to work unchanged
- Existing projects work without modification

**Testing**
Added a new test `TestMultiModuleLooseEarEjbDependency` that verifies
- Loose application XML is generated correctly
- EJB modules include dependency class directories
- Module's own classes are still included

**Before**
image

**After**
image

### DIFF
--- a/src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
@@ -126,39 +126,49 @@ public class LooseEarApplication extends LooseApplication {
     }
 
     /**
-     * Add dependency class directories so EJB modules can see classes from their dependencies
-     * @param moduleArchive
-     * @param proj
+     * Recursively collect all project dependencies (including transitive)
+     * @param project The project to collect dependencies from
+     * @param collected The set to add dependencies to
      */
+    private void collectProjectDependenciesRecursively(Project project, Set<Project> collected) {
+        // Add the project itself
+        collected.add(project)
+        
+        // Recursively collect dependencies from compileClasspath
+        if (project.configurations.findByName('compileClasspath') != null) {
+            project.configurations.compileClasspath.allDependencies.each { dep ->
+                if (dep instanceof ProjectDependency) {
+                    Project depProj = project.rootProject.findProject(dep.path)
+                    if (depProj != null && !collected.contains(depProj)) {
+                        collectProjectDependenciesRecursively(depProj, collected)
+                    }
+                }
+            }
+        }
+        
+        // Recursively collect dependencies from runtimeClasspath
+        if (project.configurations.findByName('runtimeClasspath') != null) {
+            project.configurations.runtimeClasspath.allDependencies.each { dep ->
+                if (dep instanceof ProjectDependency) {
+                    Project depProj = project.rootProject.findProject(dep.path)
+                    if (depProj != null && !collected.contains(depProj)) {
+                        collectProjectDependenciesRecursively(depProj, collected)
+                    }
+                }
+            }
+        }
+    }
+    
     private void addDependencyClassDirectories(Element moduleArchive, Project proj) {
         try {
             Set<Project> projectDependencies = new HashSet<Project>();
             
-            // Check compileClasspath for compile-time dependencies
-            if (proj.configurations.findByName('compileClasspath') != null) {
-                proj.configurations.compileClasspath.allDependencies.each { dep ->
-                    if (dep instanceof ProjectDependency) { // Ensure it's a project dependency
-                        Project depProj = proj.rootProject.findProject(dep.path)
-                        if (depProj != null) {
-                            projectDependencies.add(depProj)
-                        }
-                    }
-                }
-            }
+            // Collect all project dependencies (including transitive) recursively
+            collectProjectDependenciesRecursively(proj, projectDependencies)
+            // Remove the project itself - we only want dependencies
+            projectDependencies.remove(proj)
             
-            // Check runtimeClasspath for runtime dependencies
-            if (proj.configurations.findByName('runtimeClasspath') != null) {
-                proj.configurations.runtimeClasspath.allDependencies.each { dep ->
-                    if (dep instanceof ProjectDependency) { // Ensure it's a project dependency
-                        Project depProj = proj.rootProject.findProject(dep.path)
-                        if (depProj != null) {
-                            projectDependencies.add(depProj)
-                        }
-                    }
-                }
-            }
-            
-            logger.debug("Found ${projectDependencies.size()} project dependencies for ${proj.name}")
+            logger.debug("Found ${projectDependencies.size()} project dependencies (including transitive) for ${proj.name}")
             
             // Process each project dependency
             projectDependencies.each { dependencyProject ->
--- a/src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
@@ -172,20 +172,20 @@ public class LooseEarApplication extends LooseApplication {
                 // Add all class directories
                 dependencyProject.sourceSets.main.output.classesDirs.files.each { File classesDirectory ->
                     if (classesDirectory.exists()) {
-                        logger.debug("  Adding class dir: ${classesDirectory}")
+                        logger.debug("Adding class dir: ${classesDirectory}")
                         config.addDir(moduleArchive, classesDirectory, "/")
                     } else {
-                        logger.debug("  Skipping non-existent class dir: ${classesDirectory}")
+                        logger.debug("Skipping non-existent class dir: ${classesDirectory}")
                     }
                 }
                 
                 // Add resource directory
                 def resourcesDirectory = dependencyProject.sourceSets.main.output.resourcesDir
                 if (resourcesDirectory?.exists()) {
-                    logger.debug("  Adding resource dir: ${resourcesDirectory}")
+                    logger.debug("Adding resource dir: ${resourcesDirectory}")
                     config.addDir(moduleArchive, resourcesDirectory, "/")
                 } else {
-                    logger.debug("  No resources dir or doesn't exist: ${resourcesDirectory}")
+                    logger.debug("No resources dir or doesn't exist: ${resourcesDirectory}")
                 }
             }
         } catch (Exception e) {
--- a/src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
+++ b/src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
@@ -24,8 +24,10 @@ import org.gradle.api.logging.Logger
 import org.gradle.plugins.ear.Ear
 import org.w3c.dom.Element
 
+import java.util.jar.Manifest
+
 public class LooseEarApplication extends LooseApplication {
-    
+
     protected Task task;
     protected Logger logger;
 
@@ -61,7 +63,7 @@ public class LooseEarApplication extends LooseApplication {
             config.addFile(applicationXmlFile, "/META-INF/application.xml")
         }
     }
-    
+
     public Element addWarModule(Project proj) throws Exception {
         Element warArchive = config.addArchive("/" + proj.war.getArchiveFileName().get());
         if (proj.war.getWebAppDirectory().getAsFile().get() != null) {
@@ -98,8 +100,9 @@ public class LooseEarApplication extends LooseApplication {
         addModules(moduleArchive, proj)
         return moduleArchive;
     }
-    
+
     private void addModules(Element moduleArchive, Project proj) {
+        boolean manifestAdded = false
         for (File f : proj.jar.source.getFiles()) {
             String extension = FilenameUtils.getExtension(f.getAbsolutePath())
             switch(extension) {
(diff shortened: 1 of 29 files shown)


venmanyarun (changes requested):
as discussed, add tests

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -109,15 +112,69 @@ public class LooseEarApplication extends LooseApplication {
...
+        try {
+            def manifest = new Manifest(new FileInputStream(manifestFile))
+            String classPath = manifest.getMainAttributes().getValue("Class-Path")
+            logger.info("Processing manifest Class-Path for ${proj.name}: ${classPath}")
+            if (classPath) {
+                classPath.split(/\s+/).each { String jarName ->
+                    if (jarName && jarName.endsWith(".jar")) {
+                        String depName = jarName.replace(".jar", "")
+
+                        def depProj = proj.rootProject.findProject(":${depName}")
```
cherylking: Are we guaranteed that the dependency project name is the jarName without `.jar`? I wouldn't think so. We will be missing dependencies if they don't match right?

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -109,15 +112,69 @@ public class LooseEarApplication extends LooseApplication {
                     config.addFile(moduleArchive, f, "/WEB-INF/lib/" + f.getName());
                     break
                 case "MF":
-                    //This checks the manifest file and resource directory of the project's jar source set.
-                    //The location of the resource directory should be the same as proj.getProjectDir()/build/resources.
-                    //If the manifest file exists, it is copied to proj.getProjectDir()/build/resources/tmp/META-INF. If it does not exist, one is created there.
-                    addManifestFileWithParent(moduleArchive, f, proj.sourceSets.main.getOutput().getResourcesDir().getParentFile().getCanonicalPath())
+                    // Prefer the jar task's generated manifest (build/tmp/jar/MANIFEST.MF) which has
```
cherylking: Do we need to do something similar for Liberty Maven plugin?

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -109,15 +112,69 @@ public class LooseEarApplication extends LooseApplication {
...
+                    }
+                    manifestAdded = true
                     break
                 default:
                     break
             }
         }
+        // If no .MF file was found in the jar source set (e.g. no static MANIFEST.MF in resources),
+        // still add the jar task's generated manifest if it exists.
+        if (!manifestAdded) {
```
cherylking: Are we sure we should always do this? We don't even know if any files were added above for this module. Is this ok to do for `war` and `rar`?

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
venmanyarun: @sajeerzeji Please check by creating similar project in Liberty Maven plugin and try deploying the EAR to see issue exists in LMP

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
sajeerzeji: I did a research on this and found below cases, we can do fallback matching for these cases

1. Custom archiveBaseName - Projects using archiveBaseName = 'custom-name' produce JARs that don't match the project name
2. Versioned JARs - Projects with archiveVersion = '1.0' produce JARs like ProjectName-1.0.jar instead of ProjectName.jar
3. Classified JARs - Projects using archiveClassifier = 'client' produce JARs like ProjectName-client.jar for different variants
4. Combined customizations - Projects combining base name, version, and classifier produce complex names like custom-2.0-prod.jar

Additionally
1. Completely custom archiveFileName - Projects using archiveFileName = 'xyz-database-layer.jar' have no relationship between project name and JAR name
2. Dynamic JAR names - Projects computing JAR names at runtime using expressions like "${env}-${project.name}.jar" may not be resolvable at configuration time
3. Nested projects with custom naming - Multi-level projects (:parent:child) using custom base names like archiveBaseName = 'parent-child-combined' require path-based matching

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
sajeerzeji: Sure, will do this and will update here

sajeerzeji:
Hi @cherylking 
After reviewing the changes in this PR against the Maven plugin implementation, I identified a gap in the approach. I've reverted the previous changes and implemented a new fix as below,

Previous approach was to use manifest Class Path entries to determine dependencies, requiring users to manually declare Class-Path in their build.gradle files. This approach was reverted after cross-checking with Maven plugin implementation.

Added `addDependencyClassDirectories()` method that,
1. Scans both `compileClasspath` and `runtimeClasspath` for project dependencies
2. Automatically adds each dependency's class and resource directories to the module in loose XML
3. Works without requiring manifest Class-Path configuration


Now the plugin automatically detects project dependencies from Gradle configurations, for each EJB/JAR module, it finds all project dependencies, adds dependency class directories as `<dir>` elements in the loose XML.

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -119,5 +124,73 @@ public class LooseEarApplication extends LooseApplication {
             }
         }
     }
+
+    /**
+     * Add dependency class directories so EJB modules can see classes from their dependencies
+     * @param moduleArchive
+     * @param proj
+     */
+    private void addDependencyClassDirectories(Element moduleArchive, Project proj) {
```
venmanyarun: here, both compileClasspath and runtimeClasspath may contain the same dependencies, leading to duplicate <dir> entries in the XML.

maybe use a Set to track already-added dependencies:

src/test/groovy/io/openliberty/tools/gradle/TestMultiModuleLooseEarEjbDependency.groovy
```diff
@@ -0,0 +1,124 @@
...
+        String expression = "/archive/archive[@targetInArchive='/ejb-dependency-ejb-jar-1.0.jar']"
+        NodeList ejbModuleNodes = (NodeList) xPath.compile(expression).evaluate(inputDoc, XPathConstants.NODESET)
+        Assert.assertEquals("Should have exactly one EJB module archive", 1, ejbModuleNodes.getLength())
+
+        Node ejbModuleNode = ejbModuleNodes.item(0)
+
+        expression = "dir"
+        NodeList dirNodes = (NodeList) xPath.compile(expression).evaluate(ejbModuleNode, XPathConstants.NODESET)
+
+        Assert.assertTrue("EJB module should have at least 2 <dir> elements (own classes + dependency classes)",
```
venmanyarun: why not do an absolute test here? we should be knowing the exact count of elements in ear.xml

src/test/resources/multi-module-loose-ear-ejb-dependency-test/settings.gradle
```diff
@@ -0,0 +1,8 @@
+rootProject.name = 'ejb-dependency'
+include ':lib-jar'
```
venmanyarun, line 2: here I can see we have direct dependency with both lib-jar and ejb-jar
we should have tests that have

1. Multi-level transitive dependencies (A depends on B depends on C)
2. Mixed dependency types (project + external JARs)

venmanyarun (changes requested):

venmanyarun:
@sajeerzeji 
can you add screen recording ?

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
sajeerzeji: The project entries are getting added to a set `Set<Project> projectDependencies`

src/test/groovy/io/openliberty/tools/gradle/TestMultiModuleLooseEarEjbDependency.groovy
sajeerzeji: Absolute dirNodes count check added to the test

sajeerzeji:
> @sajeerzeji can you add screen recording ?




https://github.com/user-attachments/assets/d50d09fb-cc4a-4ec8-9f1c-caada1daa215

src/test/groovy/io/openliberty/tools/gradle/TestMultiModuleLooseEarEjbDependency.groovy
```diff
@@ -34,16 +34,8 @@ public class TestMultiModuleLooseEarEjbDependency extends AbstractIntegrationTes
...
-                   return ((!pathname.getPath().endsWith(".gradle") && 
-                            !pathname.getPath().endsWith(".gradle.kts")) ||
-                            pathname.getPath().endsWith("settings.gradle") ||
-                            pathname.getPath().endsWith("build.gradle"))
-               }
-            });
-            // Copy build.gradle
-            copyFile(new File(resourceDir, buildFilename), new File(buildDir, "build.gradle"))
+            // Copy all resources
+            FileUtils.copyDirectory(resourceDir, buildDir)
```
cherylking, line 36: Why do you think this change is ok? It will result in the build file not having the right name it many test scenarios.

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
```diff
@@ -126,39 +126,49 @@ public class LooseEarApplication extends LooseApplication {
...
-                        }
+        }
+        
+        // Recursively collect dependencies from runtimeClasspath
+        if (project.configurations.findByName('runtimeClasspath') != null) {
+            project.configurations.runtimeClasspath.allDependencies.each { dep ->
+                if (dep instanceof ProjectDependency) {
+                    Project depProj = project.rootProject.findProject(dep.path)
+                    if (depProj != null && !collected.contains(depProj)) {
+                        collectProjectDependenciesRecursively(depProj, collected)
```
cherylking, line 155: can there be a circular dependency? are you guarding against that somehow?

cherylking (changes requested):

src/test/groovy/io/openliberty/tools/gradle/TestMultiModuleLooseEarEjbDependency.groovy
sajeerzeji: I removed the unused buildFilename variable. The directory copy preserves all module build.gradle files as it is, which is the correct for multi-module tests.

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
sajeerzeji: Yes, the !collected.contains(depProj) check guards against circular dependencies. The collected Set tracks projects those already processed, so if A depends on B and B depends on A, considering the second time is skipped.

cherylking:
Looks like most of the builds are failing. So I am not going to look at the changes further until they are passing.

sajeerzeji:
> Looks like most of the builds are failing. So I am not going to look at the changes further until they are passing.

Test are passed once re-ran, the issues where network connection timeout in InstallLiberty_runtimeUrl tests.

venmanyarun (approved):

cherylking (approved):

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
cherylking: You never answered this question @sajeerzeji .

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/utils/LooseEarApplication.groovy
sajeerzeji: This was a part of previous changes, those were to use manifest Class Path entries to determine dependencies, requiring users to manually declare Class-Path in their build.gradle files. This approach was reverted after cross-checking with Maven plugin implementation.

Added addDependencyClassDirectories() method that,

Scans both compileClasspath and runtimeClasspath for project dependencies
Automatically adds each dependency's class and resource directories to the module in loose XML
Works without requiring manifest Class-Path configuration
Now the plugin automatically detects project dependencies from Gradle configurations, for each EJB/JAR module, it finds all project dependencies, adds dependency class directories as <dir> elements in the loose XML.

cherylking (approved):

PR #1085, opened 2026-05-18
changes to start hotkey reader earlier before server is fully started

venmanyarun:
Fixes https://github.com/OpenLiberty/ci.maven/issues/1638
Attaching screeen recording

https://github.com/user-attachments/assets/65995ad4-20f9-41a7-b8a6-b4db48f800a5

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -39,8 +39,9 @@ jobs:
     - name: Checkout ci.common
       uses: actions/checkout@v3
       with:
-        repository: OpenLiberty/ci.common
+        repository: venmanyarun/ci.common
         path: ci.common
+        ref: early_stop_fix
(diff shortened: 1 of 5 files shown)


cherylking (approved):
looks good..just make sure all tests pass and revert yml changes before merging

Also, testing all those hot keys in one test method seems problematic. What if dev mode finishes starting before it has looped through all the options?

venmanyarun:
> looks good..just make sure all tests pass and revert yml changes before merging
> 
> Also, testing all those hot keys in one test method seems problematic. What if dev mode finishes starting before it has looped through all the options?

Tests are flaky in ci.gradle due to the difference in time it taking to start the server( the maven rate limiting and download retries are also attributing to this) . I am excluding the tests from pipeline for now, we can enable back once we resolve the rate limiting problem

PR #1093, opened 2026-06-29
Switch both call sites from dynamic property access to NamedDomainObjectCollection.findByName(), which returns null safely, and fall back to conventional default paths when the java plugin is absent

### LINKED ISSUE #1092: libertyDev fails with "Could not get unknown property 'main' for SourceSet container" when EAR module does not apply java plugin
Running libertyDev on a multi-module project where the EAR module applies only ear and liberty (without the java plugin) fails immediately with:

```
Execution failed for task ':jims2:libertyDev'.
> Could not get unknown property 'main' for SourceSet container of type
  org.gradle.api.internal.tasks.DefaultSourceSetContainer.
```

**Steps to reproduce**

- Create a multi-module Gradle project with a WAR submodule and an EAR submodule.
- In the root build.gradle, do not apply the java plugin to all subprojects.
- In the EAR module's build.gradle, apply only ear and liberty:

```
apply plugin: 'ear'
apply plugin: 'liberty'
```

Run ./gradlew libertyDev.

The task fails immediately before the Liberty server starts

more details can be found in https://groups.io/g/openliberty/topic/119960040#

venmanyarun:
Fixes #1092 

Fix: MissingPropertyException in DevTask when Java plugin is absent (EAR projects)

DevTask.action() and DevTask.getProjectModules() both accessed sourceSets via Groovy dynamic property dispatch (project.sourceSets.main). On projects that don't apply the java plugin — such as a pure EAR module — this property doesn't exist and throws MissingPropertyException at libertyDev startup.

Changes:

- Replace project.sourceSets.main / .test with project.extensions.findByType(SourceSetContainer)?.findByName(...) at both call sites, returning null safely when the java plugin is absent.
- Fall back to conventional Gradle layout paths (src/main/java, build/classes/java/main, etc.) when sourceSets is null, matching what the Java plugin would produce anyway.
- Add integration test TestMultiModuleEarNoJavaPluginDevMode with a multi-module project (ear + war + jar) where the EAR submodule intentionally has no java plugin, reproducing the exact failing scenario.


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -1271,14 +1270,26 @@ class DevTask extends AbstractFeatureTask {
...
 
-        sourceDirectory = mainSourceSet.java.srcDirs.iterator().next()
-        testSourceDirectory = testSourceSet.java.srcDirs.iterator().next()
-        DefaultFilePropertyFactory.DefaultDirectoryVar outputDirectory = mainSourceSet.java.classesDirectory;
-        DefaultFilePropertyFactory.DefaultDirectoryVar testOutputDirectory = testSourceSet.java.classesDirectory;
-        List<File> resourceDirs = mainSourceSet.resources.srcDirs.toList();
+        if (mainSourceSet != null) {
+            sourceDirectory = mainSourceSet.java.srcDirs.iterator().next()
+        } else {
+            sourceDirectory = new File(project.projectDir, "src/main/java")
```
cherylking, line 1281: I see a lot of default paths introduced in this code update that were not there before. Are we sure it is ok? How did this work before when there was not a mainSourceSet or testSourceSet?

cherylking (changes requested):
Where is the test that uses sourceSets? I looked through all the build.gradle files and did not see it.

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
venmanyarun: In main, both call sites used direct Groovy property access:

// main branch — THROWS if java plugin absent
SourceSet mainSourceSet = project.sourceSets.main;
SourceSet testSourceSet = project.sourceSets.test;

project.sourceSets.main is not a null-returning lookup — it's a dynamic property access on NamedDomainObjectContainer that throws MissingPropertyException when the property doesn't exist. The old code had no null handling at all; it assumed the java plugin was always present.

In main, we have a test called TestMultiModuleLooseEar which was defined in a way that java plugin is added to parent build.gradle. due to this, we never had a failure

The default paths introduced by this fix are the conventional Gradle layout (src/main/java, src/test/java, build/classes/java/main, build/classes/java/test). These are safe because:

- They match what the Gradle Java plugin would produce anyway — the defaults are what sourceSets.main would resolve to in a standard project.
- They are only used when sourceSets is absent (i.e., no java plugin at all). For an EAR-only module, there are no Java sources to compile, so DevTask will simply find nothing at those paths and proceed harmlessly — the same outcome as if the EAR truly had no Java sources.

venmanyarun:
> Where is the test that uses sourceSets? I looked through all the build.gradle files and did not see it.

The test intentionally has no sourceSets block. 
That absence is the test condition — it reproduces the exact user scenario: an EAR project where the java plugin is not applied, so sourceSets does not exist on the project. 
Adding a sourceSets block would require applying the java plugin, which would make the bug disappear without the fix and would no longer test anything. 

The ear/build.gradle deliberately applies only ear + liberty plugins to prove that DevTask handles a missing SourceSetContainer gracefully. The WAR and JAR submodules in the same test project do apply the java plugin and have valid sourceSets, so both the null path (EAR) and the normal path (WAR/JAR) are exercised within the same test run.

cherylking:
The end user had a sourceSets block. I get why you have the tests without (to test the null path), but why don't we also have a test with sourceSets. There is a lot of new logic here on how you are retrieving and using the sourceSets.

venmanyarun:
> The end user had a sourceSets block. I get why you have the tests without (to test the null path), but why don't we also have a test with sourceSets. There is a lot of new logic here on how you are retrieving and using the sourceSets.

Tests added for sourceSet based projects for both devMode and deploy

src/test/resources/multi-module-loose-ear-with-sourcesets-test/war/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+        java {
+            srcDirs = ['src/main/java']
+        }
+        resources {
+            srcDirs = ['src/main/resources']
+        }
+    }
+    test {
+        java {
+            srcDirs = ['src/test/java']
```
cherylking, line 30: Did you use all the default paths here? I think so...in which case this doesn't really test much...since the code will default to these if the sourceSets are not resolved correctly by our code.

cherylking (changes requested):

src/test/resources/multi-module-loose-ear-with-sourcesets-test/war/build.gradle
venmanyarun: changed sourceSet in tests

cherylking (approved):

PR #1094, opened 2026-07-20
Support annotation processor paths and compiler args in Liberty dev mode

### LINKED ISSUE #1090: Support for annotation processor paths with transitive dependencies resolution and compiler arguments in devmode
We have added support for
1. Annotation processor paths with transitive dependencies resolution and
2. Compiler arguments

In both ci.maven and ci.common, and as there are changes in ci.common we will have to take those changes to ci.gradle as well.

References:

Issues:
https://github.com/OpenLiberty/ci.maven/issues/2021
https://github.com/OpenLiberty/ci.maven/issues/2044

PRs:
https://github.com/OpenLiberty/ci.maven/pull/2026
https://github.com/OpenLiberty/ci.maven/pull/2043
https://github.com/OpenLiberty/ci.maven/pull/2045
https://github.com/OpenLiberty/ci.common/pull/519
https://github.com/OpenLiberty/ci.common/pull/522

sajeerzeji:
Fixes #1090 
The Maven plugin recently fixed two issues where annotation processor paths (e.g. Lombok, MapStruct) and custom compiler arguments were not being forwarded to the hot reload recompiler during dev mode.
These failures do not occur in the Gradle plugin because Gradle hardcodes useBuildRecompile=true, meaning dev mode always delegates compilation to the Gradle daemon via BuildLauncher which runs the compileJava task natively. Annotation processors and compiler args are handled automatically as part of normal task execution, and JavaCompilerOptions is never reached.
But the getProjectModules() method was passing empty JavaCompilerOptions for any module in a multi-module build, which would silently drop annotation processors and compiler args if the upstream module direct recompilation path is ever implemented. This is fixed by adding a getGradleCompilerOptions() method that reads the annotation processor configuration (with full transitive dependency resolution via Gradles own dependency engine) and compiler args from each projects compileJava task, and uses it to correctly populate each upstream ProjectModule.


src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -1710,4 +1708,54 @@ class DevTask extends AbstractFeatureTask {
...
+                        options.setAnnotationProcessorPath(pathString)
+                        logger.info("Dev mode annotation processor path (compileJava.options): " + pathString)
+                    }
+                } catch (Exception e) {
+                    logger.debug("Could not resolve annotationProcessorPath on compileJava for project '" + project.name + "': " + e.getMessage())
+                }
+            } else {
+                def annotationProcessorConfig = project.configurations.findByName('annotationProcessor')
+                if (annotationProcessorConfig != null) {
+                    Set resolvedArtifacts = annotationProcessorConfig.resolvedConfiguration.resolvedArtifacts
```
venmanyarun: The fallback path calls .resolvedConfiguration.resolvedArtifacts, which forces full dependency resolution immediately. This can fail in composite builds or when called during configuration phase. Use annotationProcessorConfig.files instead

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -1710,4 +1708,54 @@ class DevTask extends AbstractFeatureTask {
         return null;
     }
 
+    /**
+     * Build a JavaCompilerOptions from the given Gradle project's compileJava task settings.
+     *
+     * @param project the Gradle project to read compiler options from
+     * @return populated JavaCompilerOptions, never null
+     */
+    protected JavaCompilerOptions getGradleCompilerOptions(Project project) {
```
venmanyarun, line 1717: here, getGradleCompilerOptions() is only wired into getProjectModules(), which only runs for upstream modules in multi-module builds. 

Both new test projects are single-module WARs — they will never enter that code path. The feature as written doesn't fix the scenario it purports to test. Wherever the root project's own JavaCompilerOptions is constructed (outside getProjectModules()), it needs the same getGradleCompilerOptions(project) call.

add a multi-module EAR integration test that exercises the upstream module's ProjectModule.getCompilerOptions()

venmanyarun (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
sajeerzeji: This is fixed

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
sajeerzeji: Added new multi-module tests

venmanyarun (approved):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
```diff
@@ -1492,11 +1492,9 @@ class DevTask extends AbstractFeatureTask {
...
         List<ProjectModule> upstreamProjects = new ArrayList<ProjectModule>();
         for (Project dependencyProject : DevTaskHelper.getAllUpstreamProjects(project)) {
-            // In Maven , there is a step to set compiler options for upstream project
-            // Gradle does not need to manually inject compiler options because
-            // we are directly calling compileJava task, which internally takes the compiler options
-            // from task definition or command line arguments
-            JavaCompilerOptions upstreamCompilerOptions = new JavaCompilerOptions();
+            // Read compiler options (annotation processor path, compiler args) from the upstream project's
+            // compileJava task so they are available for any direct javac recompilation paths.
+            JavaCompilerOptions upstreamCompilerOptions = getGradleCompilerOptions(dependencyProject)
```
cherylking, line 1497: According to a comment by Arun, `Wherever the root project's own JavaCompilerOptions is constructed (outside getProjectModules()), it needs the same getGradleCompilerOptions(project) call.` But I don't see any other call to the method getGradleCompilerOptions.

cherylking (changes requested):

src/main/groovy/io/openliberty/tools/gradle/tasks/DevTask.groovy
venmanyarun: the root project compiles via the Gradle compileJava task directly (where Gradle handles annotation processors natively), so no explicit getGradleCompilerOptions(project) call is needed for it

We should add more tests to verify this..

1.improve DevMultiModuleAnnotationProcessorTest to assert the presence of the info log "Dev mode annotation processor path (annotationProcessor config): …". 
2. In the annotation-processor multi-module project, only :jar has Lombok. Although this proves that our code is being executed,  the :war module has no annotation processor at all. There is no test that a module with an annotation processor set via compileJava.options.annotationProcessorPath . We should add a change to GreetingServlet.java to verify the WAR module still hot-reloads correctly, when lombok is configured

cherylking: Thanks for the clarification.

sajeerzeji: New assertion added in the `noCompilerOptionsResolutionWarningForJarModule` in DevMultiModuleAnnotationProcessorTest to assert the presence of the info log "Dev mode annotation processor path" log.
And in the DevMultiModuleAnnotationProcessorTest, added a new test case `modifyJavaFileInWarModuleWithLombokTriggersRecompilation` to verify that modifying GreetingServlet.java in the WAR module triggers recompilation when Lombok is configured on the WAR module itself via compileJava.options.annotationProcessorPath.

cherylking (approved):

PR #1096, opened 2026-08-03
fix: preserve backslashes in server.env values when mergeServerEnv=true

venmanyarun:
1. writeServerEnvProperties() normalised all values unconditionally — AbstractServerTask.groovy
value.replace("\\", "/") was applied to every entry in the merged file, including values read verbatim from the user's server.env. Only values that came from Gradle inline properties (env = [...] in build.gradle) should be normalised.

Fix: Added a Set<Object> normaliseKeys overload to writeServerEnvProperties(). Normalisation is applied only to keys present in that set:
copyConfigDirectory() passes Collections.emptySet() — no inline keys
setServerEnvWithAppendServerEnvHelper() passes configuredProps.keySet() — only Gradle-inline keys

2. checkServerEnvProperties() and checkEtcServerEnvProperties() mutated source files — LibertyTasks.groovy
Both methods did file.text = file.text.replace("\\", "/") on the source server.env file to extract WLP_OUTPUT_DIR — permanently rewriting the file on disk every Gradle invocation.

Fix: Introduced ServerEnvUtil.readEnvValue(File, String)— reads a single key's value line-by-line without touching the file. The extracted WLP_OUTPUT_DIR value alone still gets ?.replace("\\", "/") (it is used as a Java File path); the source file is never written.

3. Insertion order not preserved — AbstractServerTask.groovy

All Properties / HashMap usages in the merge pipeline were replaced with LinkedHashMap (convertServerEnvToProperties, combineServerEnvProperties, mergeSpecialPropsFromInstallServerEnvIfAbsent, and related helpers) so that key order from the source file is retained in the output.

Windows Test recordings

Before 

https://github.com/user-attachments/assets/4e822f33-e1bd-4573-bd8b-4506358fc070

After

https://github.com/user-attachments/assets/ae157d3c-3a18-4faa-82b5-9481149e3f3b


docs/libertyExtensions.md
```diff
@@ -42,6 +42,7 @@ The following properties are supported for server configuration.
...
 | mergeServerEnv | boolean | 3.1-M1 | Merge the server environment properties from all specified sources with the default generated `server.env` file in the target server. Conflicts are resolved with the same precedence as the replacement policy when this attribute is set to `false`. The properties specified in the `env` attribute are highest precedence, followed by the `serverEnvFile` attribute, then the `server.env` file located in the `configDirectory`, and finally the default generated `server.env` file in the target server. The default value is `false`. | No |
+| convertServerEnvPathSeparator | boolean | 3.x | When set to `true`, backslash characters (`\`) in `server.env` values are converted to forward slashes (`/`) before being written to the merged `server.env` file. This was the default behaviour in earlier versions of the plugin. The default value is `false`, which preserves values verbatim — recommended for Windows paths (e.g. `C:\my\path`) and delayed-expansion variables (e.g. `!MY_VAR!\java`). Only takes effect when `mergeServerEnv` is `true`. | No |
```
cherylking: Shouldn't this be 4.0.1 for the `Since` column?

src/main/groovy/io/openliberty/tools/gradle/extensions/ServerExtension.groovy
```diff
@@ -63,7 +63,11 @@ class ServerExtension {
     // Conflicts are resolved with the same precedence as the replacement policy when this attribute is set to `false`. 
     // The properties specified in the `env` attribute are highest precedence, followed by the `serverEnvFile` attribute, 
     // then the `server.env` file located in the `configDirectory`, and finally the default generated `server.env` file in the target server. The default value is `false`.
-    boolean mergeServerEnv = false;
+    boolean mergeServerEnv = false
```
cherylking, line 66: update copyright year

src/main/groovy/io/openliberty/tools/gradle/extensions/ServerExtension.groovy
```diff
@@ -63,7 +63,11 @@ class ServerExtension {
...
+
+    // When true, backslashes in server.env values are converted to forward-slashes before
+    // writing (legacy behaviour). Default false — values are written verbatim.
```
cherylking: Add info about this only taking affect when mergeServerEnv is true.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -869,29 +873,28 @@ abstract class AbstractServerTask extends AbstractLibertyTask {
 
     protected String handleServerEnvFileAndProperties(String serverEnvPath, String serverDirectory) {
         File envFile = new File(serverDirectory, "server.env")
-        Properties configuredProps = combineServerEnvProperties(server.env, envProjectProps);
+        // server.env and envProjectProps are Gradle-inline props (build.gradle / command line)
+        Map<String, String> configuredProps = combineServerEnvProperties(server.env ?: [:], envProjectProps)
```
cherylking: What is this syntax with `[:]`? What is being passed differently and why? Also the comment that was added is worded strange. Makes it sound like the server.env are Gradle inline props. I think the intention was to explain what envProjectProps are.

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -869,29 +873,28 @@ abstract class AbstractServerTask extends AbstractLibertyTask {
...
             if (serverEnvPath != null) {
                 logger.debug("The serverEnvFile "+ server.serverEnvFile.getCanonicalPath() + " is merged with the " + serverEnvPath + " file.")
             } else {
                 logger.debug("The serverEnvFile "+ server.serverEnvFile.getCanonicalPath() + " is merged with the " + getServerDir(project).getCanonicalPath() + " file.")
             }
-            Properties configuredServerEnvProps = convertServerEnvToProperties(server.serverEnvFile);
-            //Merge with either default server.env or with what has already been merged if
-            mergedProperties = (Properties) combineServerEnvProperties(serverEnvProps, configuredServerEnvProps);
+            Map<String, String> configuredServerEnvProps = convertServerEnvToProperties(server.serverEnvFile)
+            mergedProperties = combineServerEnvProperties(serverEnvProps, configuredServerEnvProps)
```
cherylking, line 899: Why did you remove the useful comment about what is getting merged?

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
```diff
@@ -1349,7 +1326,7 @@ abstract class AbstractServerTask extends AbstractLibertyTask {
      */
     @Internal
     protected boolean isJavaHomeSetForEnvProperties() {
-        Properties serverEnvProjectProps = combineServerEnvProperties(server.env, envProjectProps)
+        Map<String, String> serverEnvProjectProps = combineServerEnvProperties(server.env ?: [:], envProjectProps)
```
cherylking: same question about what this does

src/main/groovy/io/openliberty/tools/gradle/utils/ServerEnvUtil.groovy
```diff
@@ -0,0 +1,42 @@
...
+     * The value is returned as-is; callers are responsible for any normalisation needed.
+     */
+    public static String readEnvValue(File serverEnvFile, String key) {
+        return serverEnvFile.withReader('UTF-8') { reader ->
+            String line
+            while ((line = reader.readLine()) != null) {
+                if (!line.startsWith('#')) {
+                    String[] kv = line.split('=', 2)
+                    if (kv.length == 2 && kv[0] == key) {
+                        return kv[1]
```
cherylking: Since keys can be repeated in a server.env and the last one takes precedence, shouldn't we return the last one found?

src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
```diff
@@ -156,13 +157,10 @@ class Liberty implements Plugin<Project> {
...
         if (project.liberty.outputDir == null) {
-            Properties envProperties = new Properties()
-            //check etc/server.env and set liberty.outputDir
+            // Read WLP_OUTPUT_DIR without mutating the file. Normalise backslashes since it is a directory path.
             File serverEnvFile = new File(Liberty.getInstallDir(project), 'etc/server.env')
             if (serverEnvFile.exists()) {
-                serverEnvFile.text = serverEnvFile.text.replace("\\", "/")
-                envProperties.load(new FileInputStream(serverEnvFile))
-                Liberty.setLibertyOutputDir(project, (String) envProperties.get("WLP_OUTPUT_DIR"))
+                Liberty.setLibertyOutputDir(project, ServerEnvUtil.readEnvValue(serverEnvFile, 'WLP_OUTPUT_DIR')?.replace("\\", "/"))
```
cherylking: add a comment explaining why we do the replace here

src/main/groovy/io/openliberty/tools/gradle/LibertyTasks.groovy
```diff
@@ -15,6 +15,8 @@
  */
 package io.openliberty.tools.gradle
 
+import io.openliberty.tools.gradle.utils.ServerEnvUtil
```
cherylking, line 18: update copyright year

src/main/groovy/io/openliberty/tools/gradle/LibertyTasks.groovy
```diff
@@ -147,18 +149,13 @@ public class LibertyTasks {
...
     public void checkServerEnvProperties(ServerExtension server) {
         if (server.outputDir == null) {
-            Properties envProperties = new Properties()
-            //check server.env files and set liberty.server.outputDir
+            // Read WLP_OUTPUT_DIR without mutating the file. Normalise backslashes since it is a directory path.
             if (server.serverEnvFile != null && server.serverEnvFile.exists()) {
-                server.serverEnvFile.text = server.serverEnvFile.text.replace("\\", "/")
-                envProperties.load(new FileInputStream(server.serverEnvFile))
-                setServerOutputDir(server, (String) envProperties.get("WLP_OUTPUT_DIR"))
+                setServerOutputDir(server, ServerEnvUtil.readEnvValue(server.serverEnvFile, 'WLP_OUTPUT_DIR')?.replace("\\", "/"))
```
cherylking: Why are we doing this in multiple places? Perhaps add a utility method to the new ServerEnvUtil class that does this with one method call. Or pass a flag to the readEnvValue telling it to replace backslashes with forward slashes. And again explain why.

cherylking (changes requested):

docs/libertyExtensions.md
venmanyarun: Fixed

src/main/groovy/io/openliberty/tools/gradle/extensions/ServerExtension.groovy
venmanyarun: Fixed

src/main/groovy/io/openliberty/tools/gradle/extensions/ServerExtension.groovy
venmanyarun: comment extended to explicitly state "Only takes effect when mergeServerEnv is true."

src/main/groovy/io/openliberty/tools/gradle/utils/ServerEnvUtil.groovy
venmanyarun: Fixed

readEnvValue now scans the entire file and returns the last occurrence of the key (matching Liberty server behaviour when a key appears more than once)

src/main/groovy/io/openliberty/tools/gradle/LibertyTasks.groovy
venmanyarun: Fixed

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
venmanyarun: [:] in Groovy is an empty map literal. The original code server.env ?: [:] meant: "use server.env if it is non-null, otherwise fall back to an empty map."

This guard was unnecessary and has already been removed in the current code since server.env is declared as:

Properties env = new Properties() as it wont be null

Updated comment as well

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
venmanyarun: Added back updated comments

src/main/groovy/io/openliberty/tools/gradle/tasks/AbstractServerTask.groovy
venmanyarun: combineServerEnvProperties merges the two maps with projectProperties last (so they win on conflict), so I think empty array initialization is not needed here as well

src/main/groovy/io/openliberty/tools/gradle/LibertyTasks.groovy
venmanyarun: Moved to a utility

src/main/groovy/io/openliberty/tools/gradle/Liberty.groovy
venmanyarun: comment added in the new utility and replaced this call with the utility call

cherylking (approved):
changes lgtm

PR #1099, opened 2026-08-28
Fix dev mode test scope in multi-module Gradle projects

### LINKED ISSUE #1098: LGP Multi-Module Dev Mode Test Bug
In a multi-module Gradle project where several submodules are independently Liberty-configured (e.g. :A, :B, :C), running` gradle :A:libertyDev` and then pressing Enter to trigger tests causes dev mode to run the test task across all Liberty submodules — not just :A. Because :B and :C have not been started, their integration tests fail, and depending on task execution order, tests for :A may not run at all.

The equivalent Maven command (`mvn io.openliberty.tools:liberty-maven-plugin:dev -pl :A -am`) correctly scopes tests to module A and its compile-time dependencies only.

## Steps to Reproduce

- Create a Gradle multi-module project with a root settings.gradle that includes three subprojects: A, B, and C.
- Apply the Liberty Gradle Plugin to all three subprojects and configure a Liberty server in each.
- Start dev mode for module A only: 
` gradle :A:libertyDev`
- Once the server starts and the "Liberty is running in dev mode" message appears, press Enter to trigger on-demand tests.

## Actual Behavior
Dev mode invokes cleanTest test against the Gradle root project. Gradle expands the unqualified task names to :A:cleanTest :A:test :B:cleanTest :B:test :C:cleanTest :C:test. The tests for :B and :C fail immediately because their Liberty servers are not running. If Gradle executes :B:test or :C:test before :A:test, the build may abort before :A's tests have a chance to run.

## Expected Behavior
Pressing Enter while running `gradle :A:libertyDev` should execute `:A:cleanTest :A:test` at the minimum; however, if are to keep the test behavior consistent with the Liberty Maven plugin's current behavior, it would need to run the tests for A and its dependencies. For example, if A depends on D, and D depends on E, and they all had tests and were configured to run them, the behavior should be that tests for A, D, and E should run.

venmanyarun:
Fixes #1098 

New test class: DevModeMultiModuleScopedTestsTest
New test resource: multi-module-indep-liberty-test

Three completely independent Liberty WAR submodules — moduleA, moduleB, moduleC — each with their own Liberty server configured. None depend on each other at compile time.

Test | What it checks
-- | --
moduleADevModeStartsSuccessfully | :moduleA:libertyDev starts and Liberty reaches dev mode
onDemandTestsRunOnlyForModuleA | After pressing Enter, :moduleA: tasks appear in the log and :moduleB:test / :moduleC:test do not


venmanyarun:
Tested before and after using 3 modules

Before fixes, we can see that there are tasks run for all 3 modules even though we have only one module running in dev mode. Due to moduleB not started, its integration tests are failing

https://github.com/user-attachments/assets/165829d3-034c-4272-b20c-f927a1bd392b

After fixes, only moduelA related tests are running

https://github.com/user-attachments/assets/60ce39fe-7b37-4cba-9045-62494450f6c9

src/main/groovy/io/openliberty/tools/gradle/utils/DevTaskHelper.groovy
```diff
@@ -84,6 +84,11 @@ public class DevTaskHelper {
 
         for (Iterator<Configuration> iter = project.getConfigurations().iterator(); iter.hasNext();) {
             Configuration element = iter.next();
+            // Skip test-scoped configurations so test-only project deps are not treated as
+            // upstream compile-time modules.
+            if (element.name.toLowerCase().startsWith("test")) {
```
sajeerzeji: Could this skip a configuration which might start with `test`?

src/test/groovy/io/openliberty/tools/gradle/DevModeMultiModuleScopedTestsTest.groovy
```diff
@@ -0,0 +1,147 @@
...
+        // "Tests finished." is printed by DevUtil after runIntegrationTests returns,
+        // so once it appears the test invocation is fully complete — any moduleB or
+        // moduleC task output would already be in the log if the bug were present.
+        assertTrue(":moduleA:test must run after pressing Enter",
+                verifyLogMessage(60000, ":moduleA:test"))
+        assertTrue("Tests finished. must appear confirming the run is complete",
+                verifyLogMessage(60000, "Tests finished."))
+
+        // Now that the run is done, check scope: only moduleA tasks must have appeared.
+        assertFalse(":moduleB:test must NOT be in log",
```
sajeerzeji, line 85: assertTrue have 1 minute timeout and False have 1 second, is this correct way to do here?

src/test/resources/multi-module-indep-liberty-test/moduleA/src/main/liberty/config/server.env
sajeerzeji, line 1: Shouldn't we need server.env and server.xml in ModuleB and ModuleC? I don't think we need, but just confirming.

src/test/groovy/io/openliberty/tools/gradle/DevModeMultiModuleScopedTestsTest.groovy
venmanyarun: this is intentional and correct. The 1-minute wait on assertTrue is needed because the test run takes real time. The 1-second on assertFalse is deliberate because we've already confirmed Tests finished. — the log is complete, so we just do a fast scan

src/test/resources/multi-module-indep-liberty-test/moduleA/src/main/liberty/config/server.env
venmanyarun: Not needed as we are not starting these modules anyway

src/main/groovy/io/openliberty/tools/gradle/utils/DevTaskHelper.groovy
venmanyarun: Fixed.

replaced startsWith("test")
  with an explicit set of the 8 standard Gradle test configuration
  names (testImplementation, testCompileClasspath, etc.) so that
  custom configurations starting with "test" are not skipped

cherylking (approved):

PR #1103, opened 2026-09-02
Fix devc using getRootDir() instead of getProjectDir() in multi-module builds

### LINKED ISSUE #1102: devc resolves Containerfile/Dockerfile against project root instead of subproject directory in multi-module builds
### Description:

In a multi-module Gradle project, running libertyDevc on a submodule fails to find the default Containerfile/Dockerfile and also incorrectly resolves relative --containerfile / --containerBuildContext CLI paths. Both problems share the same root cause: project.getRootDir() is used where project.getProjectDir() is required.

### Root Cause

DevUtil resolves the default container file against the projectDirectory parameter passed to its constructor:
```

// DevUtil.java
File defaultDockerFile        = new File(projectDirectory, "Dockerfile");
File userDefaultContainerFile = new File(projectDirectory, "Containerfile");
```

The contract requires projectDirectory to be the individual submodule's base directory. Maven honours this by passing project.getBasedir(). The Gradle plugin breaks it in two places:

1. DevTaskUtil constructor call — always wrong for submodules
```

// DevTask.groovy (before fix)
this.util = new DevTaskUtil(
    ...,
    project.getRootDir(),   // ← should be project.getProjectDir()
    ...
);
```

project.getRootDir() always returns the root of the multi-module build, regardless of which subproject the task runs in. This means DevUtil looks for Containerfile/Dockerfile in the build root instead of the submodule directory.

2. convertParameterToCanonicalFile() — relative CLI paths also resolve wrong
```

// DevTask.groovy (before fix)
result = new File(project.getRootDir(), relativeOrAbsolutePath).getCanonicalFile();
//                 ^^^^^^^^^^^^^^^^^^^^^^
//  Should be project.getProjectDir()
```

When a user passes a relative path via --containerfile ./Containerfile or --containerBuildContext ./ctx, this method resolves it against the root project directory rather than the submodule, producing an incorrect absolute path.

### Steps to Reproduce

1. Create a multi-module Gradle project with at least two submodules (e.g. ear/, war/).
2. Place a Containerfile inside one submodule directory (e.g. war/Containerfile).
3. Run `./gradlew :war:libertyDevc` from the root.
4. Dev mode fails to detect the Containerfile because it searches <root>/Containerfile instead of <root>/war/Containerfile.
5. The same failure occurs when passing a relative path explicitly:
`./gradlew :war:libertyDevc --containerfile Containerfile`

Related: [OpenLiberty/ci.common#536 ](https://github.com/OpenLiberty/ci.common/issues/536)(Issue 3)

venmanyarun:
Fixes #1102 

In a multi-module Gradle project, libertyDevc failed to find the default Containerfile/Dockerfile in a subproject directory because two places in DevTask.groovy used project.getRootDir() where project.getProjectDir() is required.

Tests for container mode success

https://github.com/user-attachments/assets/51792264-82db-4dd5-91a0-79894d7522e8

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -71,7 +71,7 @@ jobs:
     - name: Run tests with Gradle on Ubuntu
       timeout-minutes: 120
       run:
-        ./gradlew clean install check -P"test.exclude"="**/DevContainerTest*,**/DevEarlyQuitTest*" -Druntime=${{ matrix.RUNTIME }} -DruntimeVersion="${{ matrix.RUNTIME_VERSION }}" --stacktrace --info --warning-mode=all
+        ./gradlew clean install check -P"test.exclude"="**/DevContainerTest*,**/DevEarlyQuitTest*,**/DevcMultiModuleContainerfileTest*" -Druntime=${{ matrix.RUNTIME }} -DruntimeVersion="${{ matrix.RUNTIME_VERSION }}" --stacktrace --info --warning-mode=all
     # Copy build reports and upload artifact if build failed
     - name: Copy build/report/tests/test for upload
       if: ${{ failure() }}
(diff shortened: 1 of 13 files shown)


cherylking (approved):
lgtm

PR #1104, opened 2026-09-02
Add devc multimodule container IT and workflow support

venmanyarun:
Test PR for https://github.com/OpenLiberty/ci.common/pull/539

Add devc multimodule container IT and workflow support

Add DevcMultiModuleContainerfileTest to verify container name uniqueness
and port assignment across modules running concurrently in dev container mode.

All dev container IT's are success

image

### DIFF
--- a/src/test/groovy/io/openliberty/tools/gradle/DevcMultiModuleContainerNameTest.groovy
+++ b/src/test/groovy/io/openliberty/tools/gradle/DevcMultiModuleContainerNameTest.groovy
@@ -1,0 +1,176 @@
+/*
+ * (C) Copyright IBM Corporation 2026.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


src/test/groovy/io/openliberty/tools/gradle/DevcMultiModuleContainerNameTest.groovy
```diff
@@ -0,0 +1,176 @@
...
+import org.junit.AfterClass
+import org.junit.BeforeClass
+import org.junit.Test
+
+/**
+ * Integration test for ci.common Issue 1 (container name collision) when multiple
+ * independently Liberty-configured Gradle submodules start with container dev mode.
+ *
+ * <p>This test starts container dev mode on two independent WAR submodules
+ * (modulea and moduleb) sequentially, and verifies that the container name written
```
sajeerzeji, line 37: As the modulea and moduleb needs to run sequentially, we need to make sure that modulea runs first right? Should we use an annotation `@FixMethodOrder(MethodSorters.NAME_ASCENDING)` here which is being used in some other tests too, for example InstallDirSubProject, InstallFeature_acceptLicense, etc?

sajeerzeji: As discussed, the sequential execution is not needed here. I am resolving this comment.

sajeerzeji (approved):

cherylking (approved):

